PR #8, opened 2017-02-24
Fixed #7: Docker-based development of Rig.

### LINKED ISSUE #7: Add support for docker based development and compilation
In order to streamline contribution let's make Rig development more accessible to developers by enabling some development/testing via Docker.

grayside:
As noted in the docker-compose file, I was unable to get the dynamic linking of -cgo compilation to work in the container. However, it does produce a working binary that can do things, I'm just not sure what it fails at right now.

`rig dashboard` and `rig start` seemed to work fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.7-alpine
+
+RUN apk add --no-cache \
+  ca-certificates \
+  git \
+  gcc \
+  musl-dev \
+  && go get github.com/tools/godep \
+  && go get github.com/mitchellh/gox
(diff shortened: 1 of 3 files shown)


grayside:
This fixes #7.

febbraro:
@grayside Do you think we should remove the cgo compilation for resolving the dnsdock container by domain name and instead inspect the container for IP address and perform the lookup that way?

grayside:
@febbraro If that removes our only dependency on cgo I think it should help avoid some real headaches. Without it you could automate the builds which seems nice.

febbraro:
@grayside ok, lets get this merged and I'll open another issue/PR around removing the cgo dependency.

Issue #5, opened 2017-02-24
Output all executed commands when --verbose and explain


grayside:
When rig shells out to execute a command, we have a teaching moment to explain some piece of the stack. Output the command and provide a one-line explanation of what it does when --verbose is used.

This follows #4 for use of the --verbose flag.

grayside:
After further thinking about this, I'm inclined to formalizing it with a separate flag entirely, `--learning`. Don't include the verbose output necessarily, do have yellow output or the like outputting the commands with messages on their use.

Commands that should show up here might be run with a StreamCommandLearning() method which has a final message parameter explaining what it does.

PR #9, opened 2017-02-24
Add --verbose flag and move some startup routine output to it.

### LINKED ISSUE #4: Create a global --verbose flag and use with rig start
`rig start` provides a lot of useful information, but it seems about twice a verbose as it needs to be. We still need some output so users have a sense of what it's doing and why it's taking 5-15seconds, but we can pull some of that information out into verbose output.

1. Add a --verbose global flag
2. Move some start output to use it.

grayside:
This reduces the amount of noise produced by default when running `rig start`. If you want to see all the details, there is still `rig --verbose start`, or setting `RIG_VERBOSE=1` as an environment variable. It might be I was too aggressive, worth reviewing.

One consideration, I'm thinking we might want a second level of verbosity within which to print out the commands ~ `out.Learning.Println(cmd.Path, cmd.Args)`. That piece is captured in #5.

This fixes #4.


grayside:
Hm, that's a bit tricky. The current change just switches the commands to output only if verbose is set, but does not address alternate processing if it isn't set. There are a couple different ways we could address this:

### Option 1: Alter StreamCommand()

We add a second command parameter to StreamCommand to run if verbose mode is active, and let the caller construct different behaviors depending on the required verbosity. I don't like this as it limits the logic to the code we are executing.

```
StreamCommand(cmd *exec.Cmd, verboseCmd *exec.Cmd)
```

### Option 2: Capture command output and return

Current StreamCommand runs the command and returns the error. If we switch to use cmd.Output() instead we can return the output or the error. This means that the caller can evaluate the current state of verbosity and process what to exec or output on it's own.

I think this one is better, but a bit more involved.

grayside:
Okay, pushed some changes.

* Moved logging setup to a separate file.
* Did some go fmt cleanup
* Adds StreamCommandForce to sidestep the verbosity checking while still using the "exec API" layer we're trying to piece together.
* status command checks verboseMode to decide which command to run.

cli/util/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	var verboseWriter = ioutil.Discard
+	if verbose {
+		verboseWriter = os.Stdout
+	}
+	logger = &RigLogger{
+		Info:    		log.New(os.Stdout, color.BlueString("[INFO] "), 0),
+		Warning: 		log.New(os.Stdout, color.YellowString("[WARN] "), 0),
+		Error:   		log.New(os.Stderr, color.RedString("[ERROR] "), 0),
+		Verbose: 		log.New(verboseWriter, "[VERBOSE] ", 0),
+		IsVerbose: 	verbose,
```
grayside: go fmt

cli/util/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func Logger() *RigLogger {
+	if logger == nil {
+		LoggerInit(false)
```
grayside: Why passing in false here?

febbraro: Well, I can;t imagine a case where the logger is nil, but just in case it is, we dont have access to the cli.Context to see if the GlobalFlag verbose is set. So, in lieu of information, lets use default behavior, which is not verbose.  Does that make sense?  I can also log a warning message about it. :)

Issue #10, opened 2017-02-28
Remove OSX dependency on cgo got go build


febbraro:
The OSX build of rig uses the `-cgo` flag so that it can use the mative OSX DNS resolution. This, currently, is simply so that I can resolve the DNS name dnsdock.outrigger.vm when listing the DNS records.  We can also accomplish this by simply getting the IP address of that container an looking it up that way. It will make the build process able to be more automated via Docker.

febbraro:
Handled in #11

PR #11, opened 2017-02-28
Removed the need for cgo compilation flag

febbraro:
Following up on #7 feedback

### DIFF
--- a/dns-records.go
+++ b/dns-records.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os/exec"
+	"strings"
 
 	"github.com/bitly/go-simplejson"
 	"github.com/urfave/cli"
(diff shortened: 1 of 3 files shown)


febbraro:
For #10

dns-records.go
```diff
@@ -42,25 +44,29 @@ func (cmd *DnsRecords) Run(c *cli.Context) error {
...
-		defer response.Body.Close()
-		if body, readErr := ioutil.ReadAll(response.Body); readErr == nil {
-			if js, jsonErr := simplejson.NewJson(body); jsonErr == nil {
-				dnsdockMap, _ := js.Map()
-				records := []map[string]interface{}{}
-				for id, value := range dnsdockMap {
-					record := value.(map[string]interface{})
-					record["Id"] = id
-					records = append(records, record)
+	if ip, ipErr := exec.Command("docker", "inspect", "--format", "{{.NetworkSettings.IPAddress}}", "dnsdock").Output(); ipErr == nil {
```
grayside: Is it worth making a utility function for this?

febbraro: We only need it in one place for now, so lets leave it. If we need it again we can undertake a larger refactoring, or decide if we need to move back to compilation with `cgo`

PR #16, opened 2017-03-08
Handle versions more broadly to account for the new Docker versioning scheme

febbraro:
Previously, versions were only numbers (1.13.1) and not padded.  Now, numbers are padded and with letters (17.03.0-ce).

Machine creation fails b/c it does not create the proper boot2docker URL so the ISO is not found.

### DIFF
--- a/cli/util/util.go
+++ b/cli/util/util.go
@@ -62,10 +62,14 @@ func AskYesNo(question string) bool {
 
 }
 
-func GetCurrentDockerVersion() *version.Version {
+func GetRawCurrentDockerVersion() string {
 	output, _ := exec.Command("docker", "--version").Output()
-	re := regexp.MustCompile("Docker version ([\\d|\\.]+)")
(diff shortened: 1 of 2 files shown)


cli/util/util.go
```diff
@@ -62,10 +62,14 @@ func AskYesNo(question string) bool {
 
 }
 
-func GetCurrentDockerVersion() *version.Version {
+func GetRawCurrentDockerVersion() string {
 	output, _ := exec.Command("docker", "--version").Output()
-	re := regexp.MustCompile("Docker version ([\\d|\\.]+)")
-	versionNumber := re.FindAllStringSubmatch(string(output), -1)[0][1]
+	re := regexp.MustCompile("Docker version (.*),")
```
grayside: Was the previous regex used for validation? Seems we've removed all constraints from versioning. Maybe it should still exclude spaces?

febbraro: The format is "Docker version [theversionnumber], build [buildhash]" so feel comfortable (for now) grabbing everything before the space.

PR #23, opened 2017-03-21
Removed $HOME from volume mount

febbraro:
Turns out it works just fine removing home, that is only needed if we want to use TLS, which, for now, we don't need to do.


febbraro:
Joanna is working on a facelift for the Dashboard frontend too. :)

PR #32, opened 2017-05-17
Provide Unison sync support for projects

febbraro:
This PR provides 2 commands, `project sync:start` and `project sync:stop`.

These commands will discover the volume name in the following order of fallback.

1. Argument passed to the command
1. Configuration in local project `.outrigger.yml` file (format of config to follow)
1. Inspecting the `docker-compose.yml` file (we need to figure out if this file should be configurable, I think it will have to be)
1. The name of the current directory

The sync command has configuration that can be provided in the `.outrigger.yml` file in he following format.

```
version: "1.0"
sync:
  volume: project-sync
  ignores:
    - "Name .gitignore"
    - "Path node_modules"
```

The ignores follow the unison ignore patterns. http://www.cis.upenn.edu/~bcpierce/unison/download/releases/stable/unison-manual.html#ignore


brew-publish.sh
```diff
@@ -72,6 +72,8 @@ class Rig < Formula
   depends_on "docker-machine"
   depends_on "docker-compose"
   depends_on "docker-machine-nfs"
+  depends_on "unison"
+  depends_on "eugenmayer/dockersync/unox"
```
grayside: I thought we were avoiding a dependency on docker-sync?

cli/commands/project.go
```diff
@@ -8,7 +8,7 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/phase2/rig/cli/commands/project"
+  "github.com/phase2/rig/cli/commands/project"
```
grayside: fmt error?

cli/commands/project.go
```diff
@@ -28,7 +28,10 @@ func (cmd *Project) Commands() cli.Command {
 		Subcommands: cmd.GetScriptsAsSubcommands(project.GetConfigPath()),
 	}
 
-	return command
+	syncStart := ProjectSync{}
```
grayside: Might be slightly more clear to name this `syncDefinition`

cli/commands/project/project-config.go
```diff
@@ -18,8 +18,14 @@ type ProjectScript struct {
 	Run         []string
 }
 
+type ProjectSync struct {
+	Volume    string
+	Ignore 		[]string
```
grayside: fmt?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
+package commands
```
grayside: Should we maybe group all the project things to the project sub-package?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+type Volume struct {
+  External bool
+}
+
+const UNISON_PORT=5000
+const MAX_WATCHES="100000"
+
+func (cmd *ProjectSync) Commands() []cli.Command {
+  start := cli.Command{
+    Name:        "sync:start",
```
grayside: Is it worth having `sync` as an alias?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+// Start the unison sync process
+func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
+  project.ConfigInit()
+  config := project.GetProjectConfigFromFile(project.GetConfigPath())
+  volumeName := cmd.GetVolumeName(ctx, config)
+  cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
+
+  // Ensure the processes can handle a large number of watches
+  if err := cmd.machine.SetSysctl("fs.inotify.max_user_watches", MAX_WATCHES); err != nil {
+    cmd.out.Error.Fatalf("Error configuring watches on Docker Machine: %v", err)
```
grayside: Let's make it a bit more clear that a watch is filesystem related, and in case sync is wrapped up in other commands, let's prefix the error with the command? I think we should take the latter up as a more common pattern.

E.g.,

```
cmd.out.Error.Fatalf("%s: Error configuring file watches on Docker Machine: %v", cmd.Name, err)
```

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+  cmd.out.Info.Printf("Starting sync volume: %s", volumeName)
+  exec.Command("docker","volume", "create", volumeName).Run()
+
+  cmd.out.Info.Println("Starting unison container")
+  exec.Command("docker","container", "stop", volumeName).Run()
+  err := exec.Command("docker","container", "run", "--detach", "--rm",
+    "-v", fmt.Sprintf("%s:/unison", volumeName),
+    "-e", "UNISON_DIR=/unison",
+    "-l", fmt.Sprintf("com.dnsdock.name=%s", volumeName),
+    "-l", "com.dnsdock.image=volume.outrigger",
```
grayside: I haven't seen a period inside a dnsdock image designation before. Is that allowed?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+    "--name", volumeName,
+    "outrigger/unison:latest",
```
grayside: Given unison needs the client and server on a fairly exact version, should we add a very specific Docker Tag to our image and then derive the tag based on the detected version available in the local system? That's a good opportunity to catch an error and provide guidance.

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+  } else {
+    cmd.out.Error.Println(err)
+  }
+
+  cmd.out.Error.Fatal("Unable to determine a name for the sync volume")
+  return ""
+}
+
+// Load the proper compose file
+func (cmd *ProjectSync) LoadComposeFile() (*ComposeFile, error) {
```
grayside: Custom parsing the compose file means any of the docker-compose magic such as variable handling will not be available. Would it be worth seeing if we can add a dependency on a golang library for this? I believe docker/libcompose is deprecated but maybe there's something else we can use?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,235 @@
...
+  yamlFile, err := ioutil.ReadFile("./docker-compose.yml")
```
grayside: So this "fallback" doesn't work for the build container?

grayside: Should we respect the COMPOSE_FILE env var, or is that just helpful enough to be confusing?

brew-publish.sh
febbraro: It's not really dockersync, it is just a utility that does the fswatching that he put in a brew tap called dockersync. We are not actually using dockersync

cli/commands/project.go
febbraro: Looks like a tab vs/ spaces issue. Please tell me you are setup to use spaces? 😄

grayside: Should be spaces, but mostly I just run go fmt and commit. Should I study that more?

cli/commands/project.go
febbraro: Agreed, should change.

cli/commands/project/project-config.go
febbraro: will do

cli/commands/project_sync.go
febbraro: There is really no concept of sub packages. They are all "Standalone" I did this previously, but since there is a BaseCommand we extend, I then got an error with cyclical imports. It was a mess, but no way around it. Since there are not subpackages it seems like it would just be easier if project_config.go was just moved into the main commands package.

cli/commands/project_sync.go
febbraro: I can take it or leave it. Not a bad idea. I'll add it in.

cli/commands/project_sync.go
febbraro: Yeah it is. Turns out we dont ever use that domain name though, so really it can just go away. I inspect the containers and use the IP address due to that cgo issue we rolled back

cli/commands/project_sync.go
febbraro: Good idea. I'll see what I can do.

cli/commands/project_sync.go
febbraro: I tried. Non of the libraries include either the most recent version of compose (v3) or the "overriding rules" that come from multiple files. They just dont have compose file parsing such that it is a library to use at the moment. Happy to transition over to one when it exists though.

cli/commands/project_sync.go
febbraro: Well, it doesn't inspect the build.yml file, but both of them need the same exact `volumes` spec in their files.

febbraro: Maybe confusing. Also if we do the docker-compose.overrride.yml thing as our "Dev" env we'd actually need to read that file instead.

cli/commands/project_sync.go
grayside: Is there some benefit to surfacing a domain to developers, e.g., for troubleshooting purposes?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,234 @@
...
+	exec.Command("rm", "-f", logFile).Run()
+
+	unisonArgs := []string{
+		".",
+		fmt.Sprintf("socket://%s:%d/", ip, UNISON_PORT),
+		"-auto", "-batch", "-silent", "-contactquietly",
+		"-repeat", "watch",
+		"-prefer", ".",
+		"-logfile", logFile,
+		"-ignore", "Name .git",
```
mike-potter: Should this be "Name .git" or "Path .git"?

mike-potter: Nevermind, "Name .git" is correct.

cli/commands/project.go
febbraro: changed

cli/commands/project/project-config.go
febbraro: formatted

cli/commands/project_sync.go
febbraro: added sync alias

cli/commands/project_sync.go
febbraro: I guess it is easier if folks want to point a local run unison process at it for some reason, but I dunno, it maybe just confusing, or it may just help people understand what thecontainer is even there for, Thoughts?

cli/commands/project_sync.go
febbraro: done

cli/commands/project_sync.go
```diff
@@ -0,0 +1,249 @@
...
```
grayside: Shouldn't there be some processing to combine this with the sync configuration for ignores I see in the project.ProjectSync struct?

cli/commands/project_sync.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// 4. Use local dir for the volume name
+	if dir, err := os.Getwd(); err == nil {
+		var _, folder = path.Split(dir)
+		return folder
+	} else {
+		cmd.out.Error.Println(err)
+	}
+
+	cmd.out.Error.Fatal("Unable to determine a name for the sync volume")
```
grayside: This skates by review, but once we have a docs page for this we should link to it here.

PR #33, opened 2017-05-17
Travis Support

grayside:
This creates a baseline Travis integration configuration and adds a simple script to see if any of the files are not properly formatted.


grayside:
Looks like this will fail for lack of having godep installed. In theory travis can run that automatically but our use of Godeps in a subdirectory confuses things.

grayside:
This seems to be working successfully on Linux but not OSX. Can we remove the OSX builds?

PR #34, opened 2017-05-19
DEVTOOLS-229: Adding project create command

tekante:


cli/commands/project_create.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// keep passed in args as distinct elements or they will be treated as
+	// a single argument containing spaces when the container gets them
+	args := []string{
+		"run",
+		"--rm",
+		"-it",
+		"-v", fmt.Sprintf("%s:/generated", cwd),
+		image,
+		"yo", "--no-insight",
```
tekante: Should we add --skip-install as a default (and for both --no-insight and --skip-install should there be flags to rig project create to suppress those from passing in).

grayside: @mike-potter  baked those in to the outrigger/yeoman base image. The version I'm guessing you are experimenting with is what we currently have available from private bitbucket, but we want to use a public one for this which will have a few minor differences.

cli/commands/project_create.go
```diff
@@ -0,0 +1,71 @@
...
+
+type ProjectCreate struct {
+	BaseCommand
+}
+
+func (cmd *ProjectCreate) Commands() []cli.Command {
+	create := cli.Command{
+		Name:        "create",
+		Aliases:     []string{},
+		Usage:       "Run a yeoman generator in a container to create project configuration files",
```
grayside: Can we change this to something like "Run a code generator to create new project scaffolding."

There's code as well as configuration in generator output, and I'm not sure we need to specify yeoman/container in the short description.

cli/commands/project_create.go
```diff
@@ -0,0 +1,71 @@
...
+		ArgsUsage:   "[optional type] [optional args]",
+		Description: "The type is the generator to run with args passed to that generator. If using flag arguments use -- before specifying type and arguments.",
```
grayside: Should we add any further description based on special options of the default generator?

cli/commands/project_create.go
mike-potter: Yeah, you don't even need the "yo" command.  Just pass any specified "type" and any additional specified "args".  The container will run yo directly and will add the --skip-install and --no-insight options.

Basically, rig doesn't have to know anything about this being "yeoman".

cli/commands/project_create.go
tekante: Did you have particular thoughts about options that should be called out?

cli/commands/project_create.go
tekante: Updated to match expected input of the image produced from https://github.com/phase2/outrigger-generator

cli/commands/project_create.go
tekante: Updated this to: "Run a code generator to generate scaffolding for a new project."

Issue #40, opened 2017-05-26
Configured scripts without a run section are silently dropped


grayside:
If you configure a script such as:

```
scripts:
  ninja:
    description: Ninja vanish
```
or
```
scripts:
  ninja:
    run:
```
the script does not register with rig project.

There should be a verbose or warning level log explaining the incomplete script configuration.

Bonus points if we implement the check for this problem via a `rig project config` or `rig project validate` command.

febbraro:
We validate every config file when we read it, we should just add that validation there, but have a way to determine what level of severity or scrutiny we place on validation (when running a project command we want lots of scrutiny)

grayside:
Right, I'm suggesting we should have a command dedicated to that, and wondering if setting it up in that way requires changing our approach for further validation rules.

febbraro:
What is the rationale for a separate command that you have to run, instead of just reporting these problems when the project specific configuration is in play?

grayside:
Similar to the rationale of `docker-compose config -q` or `composer validate`, it means we can test the correctness/diagnose problems with the configuration before committing changes.

Tightening up the validation without regard for a separate command is the priority, but if possible we should let problems trickle up to the caller to determine what/how to report.

febbraro:
Resolved in #43. If we want a `rig project validate` or `rig project config` then lets open another issues specifically for that.

Issue #39, opened 2017-05-26
Configure a script key without any values causes a panic


grayside:
If you add a script definition to an outrigger.yml, such as:

```
scripts:
    awesome:
```
it kernel panics on running simple things like `rig project`. We should detect this state and output an error message indicating the broken configuration. Ideally this is something that would be part of a `rig project config` or `rig project validate` command.

Because I'm proposing an enhancement to fix an edge case bug, I've added both labels.

febbraro:
Resolved in #43

Issue #37, opened 2017-05-26
Inconsistent handling of alias/command conflicts


grayside:
In our current pre-release we have `rig project create`. If we have a configured script with the alias `create`, the result is inconsistent:

* The built-in command is executed on `rig project create`
* The aliased help is displayed on `rig project help create`

Since this functionality is relying on the command controller implemented in https://github.com/urfave/cli, I think this needs an upstream fix.

grayside:
Though per discussion and some of the ideas in #39 and #40, maybe we could validate the alias is non-conflicting and leave upstream to it's own devices.

febbraro:
Resolved in #43

PR #43, opened 2017-05-31
Enhanced project config validation

febbraro:
This handles (as failure cases) the validation issues reported in #37 #39 #40

### DIFF
--- a/cli/commands/project.go
+++ b/cli/commands/project.go
@@ -29,21 +29,23 @@ func (cmd *Project) Commands() []cli.Command {
 		Before:      cmd.Before,
 	}
 
-	if subcommands := cmd.GetScriptsAsSubcommands(); subcommands != nil {
-		command.Subcommands = subcommands
-	}
-
 	create := ProjectCreate{}
 	command.Subcommands = append(command.Subcommands, create.Commands()...)
 
 	sync := ProjectSync{}
 	command.Subcommands = append(command.Subcommands, sync.Commands()...)
 
+	if subcommands := cmd.GetScriptsAsSubcommands(command.Subcommands); subcommands != nil {
+		command.Subcommands = subcommands
+	}
+
 	return []cli.Command{command}
 }
 
 // Processes script configuration into formal subcommands.
-func (cmd *Project) GetScriptsAsSubcommands() []cli.Command {
+func (cmd *Project) GetScriptsAsSubcommands(otherSubcommands []cli.Command) []cli.Command {
+
+	cmd.Config.ValidateProjectScripts(otherSubcommands)
 
 	if cmd.Config.Scripts == nil {
 		return nil
--- a/cli/commands/project_config.go
+++ b/cli/commands/project_config.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/phase2/rig/cli/util"
+	"github.com/urfave/cli"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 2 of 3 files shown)


cli/commands/project.go
```diff
@@ -29,21 +29,23 @@ func (cmd *Project) Commands() []cli.Command {
 		Before:      cmd.Before,
 	}
 
-	if subcommands := cmd.GetScriptsAsSubcommands(); subcommands != nil {
-		command.Subcommands = subcommands
-	}
-
 	create := ProjectCreate{}
 	command.Subcommands = append(command.Subcommands, create.Commands()...)
```
grayside: Shouldn't this fail since you are not initializing command.Subcommands first?

cli/commands/project_config.go
```diff
@@ -80,22 +81,55 @@ func NewProjectConfigFromFile(filename string) *ProjectConfig {
...
 
-	if config.Scripts != nil {
-		for id, script := range config.Scripts {
-			if len(script.Run) > 10 {
-				util.Logger().Warning.Printf("Project script '%s' has more than 10 run items (%d). You should create a shell script to contain those.", id, len(script.Run))
+	return nil
+}
+
+// Validate the config scripts against a set of rules/norms
+func (c *ProjectConfig) ValidateProjectScripts(subcommands []cli.Command) {
```
grayside: While it wasn't described in the issues, what happens if a config file attempts to use the same script key twice?

cli/commands/project_config.go
```diff
@@ -80,22 +81,55 @@ func NewProjectConfigFromFile(filename string) *ProjectConfig {
...
+
+			// Check for an empty script
+			if script == nil {
+				logger.Error.Fatalf("Project script '%s' has no configuration", id)
+			}
+
+			// Check for scripts with conflicting aliases with existing subcommands or subcommand aliases
+			for _, subcommand := range subcommands {
+				if id == subcommand.Name {
+					logger.Error.Fatalf("Project script name '%s' conflicts with command name '%s'. Please choose a different script name", id, subcommand.Name)
```
grayside: The downside of "throwing" the error here is that if someone is in a working project directory, and runs `rig start` a broken outrigger.yml file will cause rig to bail. Or am I wrong?

That's what I was trying to get at with the idea of surfacing messages and an error state for the caller to decide what to do. I suppose we want `rig project` to choke on this, but we don't really care if a given outrigger.yml file is broken for other commands.

cli/commands/project_config.go
febbraro: I'm guessing the second one wins. Should we check keys and aliases against other scripts as well?

cli/commands/project.go
febbraro: Since it is not a pointer to a slice, it gets initialized to an empty slice by default.

https://github.com/urfave/cli/blob/master/command.go#L44

cli/commands/project_config.go
febbraro: It's difficult, but not impossible. So you want it to fail on this validations on a project run invocation, not in just any old command execution?  There is a Before action that we use to initialize the logger and the machine from the global params and maybe tie into that, just on project subcommands?

cli/commands/project_config.go
grayside: That's a nicetohave, but it does make sense to make sure we don't cross the script/alias streams as an enforced clarity practice.

cli/commands/project_config.go
grayside: I think we should do that as a follow-up, but yes, basically. I was thinking we log non-error/non-fatal things, and when we do hit one, we return that as an error and let the caller deal with it. Not sure that needs any kind of Before manipulation.

febbraro: I can easily move the validation function, but i may have to jump some hoops to get all the Subcommands at that time. Not sure if it is accessible within an executing command.

PR #50, opened 2017-06-21
Automatically pull generator and dashboard before using

### LINKED ISSUE #49: Run docker pull as part of `rig project create`

grayside:
Fixes #49

### DIFF
--- a/cli/commands/dashboard.go
+++ b/cli/commands/dashboard.go
@@ -18,13 +18,6 @@ func (cmd *Dashboard) Commands() []cli.Command {
 		{
 			Name:  "dashboard",
 			Usage: "Start Dashboard services on the docker-machine",
-			Flags: []cli.Flag{
-				cli.BoolFlag{
-					Name:   "no-update",
-					Usage:  "Prevent automatic update of dashboard.",
-					EnvVar: "RIG_DASHBOARD_NO_UPDATE",
-				},
-			},
 			Before: cmd.Before,
 			Action: cmd.Run,
 		},
@@ -34,7 +27,7 @@ func (cmd *Dashboard) Commands() []cli.Command {
 func (cmd *Dashboard) Run(ctx *cli.Context) error {
 	if cmd.machine.IsRunning() {
 		cmd.out.Info.Println("Launching Dashboard")
-		cmd.LaunchDashboard(cmd.machine, !ctx.Bool("no-update"))
+		cmd.LaunchDashboard(cmd.machine)
 	} else {
 		cmd.out.Error.Fatalf("Machine '%s' is not running.", cmd.machine.Name)
 	}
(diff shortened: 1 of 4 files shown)


cli/commands/dashboard.go
```diff
@@ -16,33 +16,58 @@ type Dashboard struct {
...
+		cmd.LaunchDashboard(cmd.machine, ctx.Bool("no-update"))
 	} else {
 		cmd.out.Error.Fatalf("Machine '%s' is not running.", cmd.machine.Name)
 	}
 
 	return nil
 }
 
-func (cmd *Dashboard) LaunchDashboard(machine Machine) {
+func (cmd *Dashboard) LaunchDashboard(machine Machine, update bool) {
```
febbraro: Is `update` confusing here. True means don;t update, right?

cli/commands/project_create.go
```diff
@@ -41,26 +49,57 @@ func (cmd *ProjectCreate) Create(ctx *cli.Context) error {
...
+	// Keep passed in args as distinct elements or they will be treated as
+	// a single argument containing spaces when the container gets them.
 	args := []string{
 		"container",
 		"run",
 		"--rm",
 		"-it",
 		"-v", fmt.Sprintf("%s:/generated", cwd),
+		"--name",
+		generator_container_name,
```
febbraro: Why are we naming the container here?

cli/commands/start.go
```diff
@@ -112,7 +112,7 @@ func (cmd *Start) Run(c *cli.Context) error {
 
 	cmd.out.Verbose.Println("Launching Dashboard...")
 	dash := Dashboard{BaseCommand{machine: cmd.machine, out: cmd.out}}
-	dash.LaunchDashboard(cmd.machine)
+	dash.LaunchDashboard(cmd.machine, true)
```
febbraro: If we update every time we start the VM, why would we want to provide the option to not update in the dashboard command. We should remove that and always update.

cli/commands/dashboard.go
grayside: Ah, I got that flipped. Will fix.

cli/commands/project_create.go
grayside: Since we name the Dashboard container. Since this might actually be done concurrently I'll remove that.

cli/commands/start.go
grayside: My thought was that all containers we use should have an auto-update with the option to suppress.

cli/commands/dashboard.go
```diff
@@ -41,8 +41,24 @@ func (cmd *Dashboard) LaunchDashboard(machine Machine) {
...
+
+	// The check for whether the image is older than 30 days is not currently used.
+	_, seconds, err := util.ImageOlderThan(image, 86400*30)
+	if err == nil {
+		cmd.out.Verbose.Printf("Local copy of the image '%s' was originally published %0.2f days ago.", image, seconds/86400)
+	}
 
+	// If there was an error it implies no previous instance of the image is available
+	// or that docker operations failed and things will likely go wrong anyway.
+	if err == nil {
```
febbraro: Can this be combined with the if err == nil above? its the same check

febbraro: Or it doesn;t matter if it exist locally or not. Just always pull it.

PR #62, opened 2017-08-15
Better sync detection and configuration via environmental variables

tekante:
This actually waits for the unison log file to stop growing before claiming that the initial sync is done.


grayside:
Would mtime be simpler/more efficient?

grayside:
The failure of the branch implies the need to run `go fmt` on the touched files.

tekante:
There doesn't appear to be any code advantage to trying to work with mtime, it's actually a little more complex due to needing to deal with the time type. It also doesn't appear to offer any speed advantages in terms of allowing a shorter pause between checks. It may be slightly more efficient but I don't have any data to know whether go is actually going to fetch all the stat data data regardless and then just push back the piece we need. Regardless, we're not checking frequently enough that I think efficiency at this scale is of concern.

I'll run stuff through gofmt and push up an update.

PR #64, opened 2017-08-16
Add go vet to travis testing [WIP]

grayside:
I've recently come across [go vet](https://godoc.org/cmd/vet), [golint](https://github.com/golang/lint) and [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) as tools for code quality. go vet is the only one suitable for a CI process.

If I were to compare the vet and lint tools to the PHP ecosystem, vet is phplint with a dash of phpmd, while lint is phpcs.


grayside:
Looks like there was a lot of direct merging into master today.

cli/commands/project_doctor.go
```diff
@@ -0,0 +1,50 @@
+package commands
```
febbraro: Was this file added mistakenly?

grayside:
Yes, that was a WIP that I forgot was not stashed. Only the .travis.yml change is needed, and it's catching a potential issue.

grayside:
Removed the project_doctor.go file.

febbraro:
I ran these checks and it came back with a bunch of errors that are not actually fixable, as best I can tell.  not sure if this should be merged if it will throw a bunch of false positives.

```bash
✘  ~/Projects/src/github.com/phase2/rig/cli   develop ●  go tool vet -v commands
vet: commands/command.go:5:2: could not import github.com/phase2/rig/cli/util (can't find import: github.com/phase2/rig/cli/util)
Checking file commands/command.go
Checking file commands/config.go
Checking file commands/dashboard.go
Checking file commands/data_backup.go
commands/data_backup.go:29: can't check non-constant format in call to Sprintf
Checking file commands/data_restore.go
Checking file commands/dns-records.go
Checking file commands/dns.go
Checking file commands/doctor.go
Checking file commands/kill.go
Checking file commands/machine.go
Checking file commands/project.go
commands/project.go:58: can't check non-constant format in call to Sprintf
commands/project.go:60: can't check non-constant format in call to Sprintf
Checking file commands/project_config.go
Checking file commands/project_create.go
Checking file commands/project_sync.go
Checking file commands/prune.go
Checking file commands/remove.go
Checking file commands/restart.go
Checking file commands/start.go
Checking file commands/status.go
Checking file commands/stop.go
Checking file commands/upgrade.go
commands/upgrade.go:29: can't check non-constant format in call to Sprintf
commands/upgrade.go:54: printf argument machineDockerVersion is pointer to invalid or unknown type
Checking file commands/watch.go
```

tekante:
This https://github.com/golang/go/issues/17571 seems to indicate that for the import error at least `go tool vet` isn't ever going to be the right thing to run. I haven't been able to convince myself that I've gotten it to do a valid scan using `go vet` in a variety of incarnations as it never seems to spit out any error messages.

grayside:
I found one thing which seemed legitimate, a case where a Fatalf() was needed (or at least preferred) but only a Fatal() was in the code. I'll do a bit more checking on this another time... came across it while trying to investigate testing generally.

grayside:
By the time anyone cares to figure this out again, it will probably need a new approach.

Issue #69, opened 2017-09-11
Add support for Desktop notifications


grayside:
###  Problem

Some of our operations take awhile to run, and others (such as rig doctor) may have value to run in the background. Integration with OSX, Linux, and Windows desktop notification systems would help surface the completion of those operations if developer focus shifts.

### Proposed Solution

* Find a cross-OS library (though newer, I like the look of https://github.com/martinlindhe/notify)
* Find a uniform way to add flags with env var options to enable/disable notifications for various rig commands.
* Have a global env var "kill switch" for people that hate notifications.
* If possible, centrally automate command success/failure messages.

From work-in-progress, the last point seems impossible, as:
* returning error states causes the framework to exit
* There is no obvious way to perform inheritance or chaining of actions other than singular declarative callbacks.

PR #70, opened 2017-09-12
Desktop notifications [WIP]

grayside:
As a precursor to evaluating this, we are discussing how to react to fail states in #78. Marking this as a WIP.

This is a first stop towards #69, the addition of dependencies to wrangle cross-OS notifications, a hidden command to facilitate developer testing, and a way to add centralized flags.

We should not merge this as-is because it enables notification flags for a bunch of commands that do not actually support notifications, I left that in here to illustrate how that would look.

I do not like configuring the notification defaults in main like that, the only idea I have left on that is to add it to the command definition struct.

Another forthcoming challenge: Using the frameworks exit codes and triggering notifications cleanly. I've started refactoring such as the following:

```
func (cmd *Start) Run(c *cli.Context) error {
  err := cmd.StartOutrigger(c)
  notify.CommandStatus(c, err == nil)
  return err
}
```

Where err is set by:
```
  cmd.out.Error.Println("Vagrant NFS mount found blocking setup of Outrigger NFS mount.")
  return cli.NewExitError("Please remove any non-Outrigger NFS mounts that begin with /Users from your /etc/exports file", 1)
```

Unfortunately, the After function provided by the framework does not work if you return an error, but I do like that this splits our controller from our API functions.

I wanted to stop here because the amount of refactoring is larger than I originally expected and I want to make sure I'm on the right track before spending more time.

**Remaining Tasks**

* [ ] Finalize the control flow and notifications opt-in approach
* [ ] Identify how, if at all, we might support rig project scripts. Doing so is a follow-up.


cli/commands/noop.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/urfave/cli"
+
+	"github.com/phase2/rig/cli/notify"
+)
+
+type Noop struct {
+	BaseCommand
+}
+
+func (cmd *Noop) Commands() []cli.Command {
```
febbraro: Was it a mistake to leave this command here?

cli/main.go
```diff
@@ -55,6 +57,34 @@ func main() {
 	app.Commands = append(app.Commands, (&commands.Watch{}).Commands()...)
 	app.Commands = append(app.Commands, (&commands.Project{}).Commands()...)
 	app.Commands = append(app.Commands, (&commands.Doctor{}).Commands()...)
+	app.Commands = append(app.Commands, (&commands.Noop{}).Commands()...)
+
+	notificationTitle := fmt.Sprintf("Outrigger CLI (rig) v%s", VERSION)
+	notify.Init(app.Name, notificationTitle, "images/logo.png", map[string]bool{
```
febbraro: Should we let the commands themselves contain the config for notification status?

cli/commands/noop.go
grayside: I left it in on purpose, but we can remove it if we want. It's a hidden command we can use to directly test notification behavior.

cli/main.go
grayside: Yes, I didn't understand how to modify the command struct until the last few days. We should definitely move this.

cli/notify/notify.go
```diff
@@ -0,0 +1,121 @@
...
+	// The name of the app for environment variables.
+	Id string
+	// Relative path to notification logo.
+	Icon string
+	// The label for the app to be displayed in notifications.
+	Label string
+}
+
+// Initialize notifications.
+func Init(id string, label string, icon string, commands map[string]bool) error {
```
febbraro: I'm thinking that it likely makes sense to make Notify a struct/object since it has config that is uses and such. Kind of like Commands themselves.

PR #73, opened 2017-10-05
Add linux bind volume support

febbraro:
Some exploration into having linux use bind volume mounts (instead of unison) so that we don't need different configs for volumes on mac & linux

### DIFF
--- a/cli/commands/project_sync.go
+++ b/cli/commands/project_sync.go
@@ -77,14 +77,12 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
 	volumeName := cmd.GetVolumeName(ctx, config)
 
 	switch platform := runtime.GOOS; platform {
-	case "darwin":
-		cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
-		cmd.StartUnisonSync(ctx, volumeName, config)
 	case "linux":
 		cmd.out.Verbose.Printf("Setting up local volume: %s", volumeName)
 		cmd.SetupBindVolume(volumeName)
 	default:
-		cmd.out.Verbose.Printf("Volume Sync not supported on: %s", platform)
+		cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
+		cmd.StartUnisonSync(ctx, volumeName, config)
 	}
 
 	return nil
--- a/cli/commands/project_sync.go
+++ b/cli/commands/project_sync.go
@@ -77,14 +77,14 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
 	volumeName := cmd.GetVolumeName(ctx, config)
 
 	switch platform := runtime.GOOS; platform {
-		case "darwin":
-			cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
-			cmd.StartUnisonSync(ctx, volumeName, config)
-		case "linux":
-			cmd.out.Verbose.Printf("Setting up local volume: %s", volumeName)
-			cmd.SetupBindVolume(volumeName)
-		default:
-			cmd.out.Verbose.Printf("Volume Sync not supported on: %s", platform)
+	case "darwin":
+		cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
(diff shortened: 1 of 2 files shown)


cli/commands/project_sync.go
```diff
@@ -74,8 +75,23 @@ func (cmd *ProjectSync) Commands() []cli.Command {
...
 
+	switch platform := runtime.GOOS; platform {
+	case "darwin":
+		cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
+		cmd.StartUnisonSync(ctx, volumeName, config)
+	case "linux":
+		cmd.out.Verbose.Printf("Setting up local volume: %s", volumeName)
+		cmd.SetupBindVolume(volumeName)
+	default:
+		cmd.out.Verbose.Printf("Volume Sync not supported on: %s", platform)
```
grayside: Does unison not work on Windows?

febbraro: Don't know, we have not tried it yet, but at a minimum the binary would likely be something like `unison.exe` so our current command would not be supported. I think windows also does cli options differently.

grayside: Ah, so currently we are in aggressive conflict with Windows. Perhaps in the recent change to generator-outrigger-drupal, I should have it use the "devcloud" options for non-Darwin, instead of for linux specifically.

febbraro:
Moved from WIP to legit PR.  Verified `sync:start` as working on Linux, and also folded in a few more compat fixes to doctor

cli/commands/project_sync.go
grayside: I'm confused what hole are we leaving Windows users in? Is it "Add Unison support on Windows" or is it "Treat Windows like Linux"?

cli/commands/project_sync.go
```diff
@@ -123,19 +141,39 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
...
+
+	cmd.out.Info.Printf("Starting local bind volume: %s", volumeName)
+	exec.Command("docker", "volume", "rm", volumeName).Run()
+
+	if workingDir, err := os.Getwd(); err == nil {
+		volumeArgs := []string{
+			"volume", "create",
+			"--opt", "type=none",
+			"--opt", fmt.Sprintf("device=%s", workingDir),
+			"--opt", "o=bind",
```
grayside: To be clear, the lack of specifying ro meant this is a read/write binding?

cli/commands/project_sync.go
```diff
@@ -123,19 +141,39 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
...
+		exec.Command("docker", volumeArgs...).Run()
+	} else {
+		cmd.out.Error.Fatalf("Error resolving the working directory for volume creation: %v", err)
+	}
 }
 
-// Start the unison sync process
 func (cmd *ProjectSync) RunStop(ctx *cli.Context) error {
+	if runtime.GOOS == "linux" {
+		cmd.out.Info.Println("No unison container to stop, using local bind volume")
```
grayside: Since we are creating the volume as an external, shouldn't this stop the volume for them?

And if so, as long as any action is taken I think logging about no-op should be verbose.

cli/commands/project_sync.go
febbraro: Windows support would be "Treat Windows like Mac"

cli/commands/project_sync.go
febbraro: Volumes are not start or stopped, they just exist. Are you suggesting we do the volume remove here in the stop?

cli/commands/project_sync.go
febbraro: Oh, I see what you mean.... let's default to what mac does and lets just use Linux as the single case.

cli/commands/project_sync.go
grayside: Nope, you are right, we should probably not do a volume remove here.

cli/commands/project_sync.go
febbraro: Are you suggesting that we explicitly mark it `:rw`

grayside: Just confirming its `rw`.

PR #78, opened 2017-10-10
Demonstrate conversion of cmd.out.Error.Fatalf to cli.NewExitError [WIP]

grayside:
cmd.out.Error.Fatalf is the Drupal equivalent to die() instead of drupal_exit(). cli.NewExitError() is the framework approach to telling the framework to bail with that error code and message as soon as it's ready.

This will help enable notifications as a practical benefit. In the notifications PR of #70, the premise is that we configure our command callbacks to be our own "controller", and that controller can intercept these returned errors and take actions before returning the error to be handled by the framework proper.

If greenlit, I will expand this branch to include most instances of the Fatalf pattern, as we don't have many (if any) cases where immediately halting the program is essential.

### DIFF
--- a/cli/commands/start.go
+++ b/cli/commands/start.go
@@ -53,6 +53,12 @@ func (cmd *Start) Commands() []cli.Command {
 }
 
 func (cmd *Start) Run(c *cli.Context) error {
+	err := cmd.StartOutrigger(c)
+	notify.CommandStatus(c, err != nil)
+	return err
+}
+
+func (cmd *Start) Outrigger() error {
 	if runtime.GOOS == "linux" {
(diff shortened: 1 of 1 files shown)


febbraro:
So without doing this, `rig` is always exiting with a 0 return code, even if we Error.Fatalf?

grayside:
No, it still exits with a status code of `1`. Without this we are:

* Unable to react to failures with a clean pattern.
* Unable to specify the error code (which can be nice if we want to have an error glossary)
* Short-circuiting anything the framework might choose to do before exiting, which currently isn't anything before checking and calling https://github.com/urfave/cli/blob/master/errors.go#L78

grayside:
We are taking the approach in #84

PR #79, opened 2017-10-11
Fix Linux compatibility of stop and restart

grayside:


cli/commands/stop.go
```diff
@@ -25,6 +25,27 @@ func (cmd *Stop) Commands() []cli.Command {
...
+	case "linux":
+		return cmd.StopMinimal()
+	default:
+		return cmd.StopOutrigger()
+	}
+}
+
+// Stop "minimal" Outrigger operations, which refers to Linux environments where
+// a virtual machine and networking is not managed by Outrigger.
+func (cmd *Stop) StopMinimal() error {
```
grayside: Playing with the idea we should be treating this as a feature detection rather than environment thing.

Issue #85, opened 2017-10-14
[Project Sync] Remove optional argument for volume name


grayside:
The presence of this argument means it is impossible to infer from the project config the name of the sync volume, which means creating tests is not possible. We can support special circumstances to override or change the volume name via config overrides as discussed in #72.

Unfortunately this would be a backwards compatibility break.

febbraro:
I'm not sure I understand why this is a problem. Is it a problem because it is an argument and not an option?  Would converting it to an option remove the problem?  Can you explain it a bit more maybe with an example? (sorry 🙀 )

One of the main reasons it is here is so that you don't need a project config file to make use of a *particular* volume.  I don't think this is in widespread usage, so it is probably ok to change how it is being used.

grayside:
1. How do we test/doctor check that a volume exists for a given project, given this can ad hoc override what the project would use?
2. How do we create a command to manage the volume or container given there is non-contextual information driving what this could be called?

The crux of this is that the project perspective of this is mostly about rig magically handling filesync with a few outrigger.yml options to tweak behavior, but the parameters and arguments on the command are about using rig as a wrapper or a set of docker run commands.

I have been thinking about whether there is value in having a `rig fs` top-level command, and allowing `rig fs sync` to be the place where this is flexibly operated. (Incidentally, this might dovetail with #6). Then `rig project sync` focus on this as a polished, config driven product can be split from the other use cases of facilitating unison sync as an ad hoc/power user function.

febbraro:
I see what you mean. I dont have particularly strong feels one way or another, and honestly I don't think that the `sync:start [volume-name]` is being used by anyone yet.

PR #86, opened 2017-10-14
project sync executed in project root directory

### LINKED ISSUE #67: rig project sync does not set working directory
All shell executions by a `rig project *` command should set the working directory to the directory of the config file.

In this way a developer could export the location of a config file in a terminal session, then run commands relevant to the project regardless of their actual working directory.

An example of this can be seen around https://github.com/phase2/rig/blob/develop/cli/commands/project.go#L89.

grayside:
Corrects the assumption in project_sync.go that commands should be executed from the current working directory. Instead, commands are executed the same as rig project scripts -- from the directory in which the configuration file is found.

* [x] If no project config is found, using current working directory as the local side of the file sync, per existing shipped behavior.
* [x] RemoveLogFile resolution: inline the logic or make a stand-alone API function.

Fixes #67


grayside:
This needs more testing for execution when there isn't any project configuration.

grayside:
The change as implemented removes support for the current working directory as a fallback to their being no configuration files available, so that needs to be put back.

febbraro:
So you are saying that before this PR the root for the sync dir was whatever dir you were in when you ran the command, and not the directory that the outrigger.yml file is in?

Now with this PR it will only work if there is an outrigger.yml and use that file's directory, but we also need to account for the current working directory if that project config file doesn't exist?

Also, with the issue you linked, recursive project config discovery, that seems kind of problematic. Maybe the project config file need to specify the sync dir, and if there is no project config in the recursive heirarchy that specifies a sync dir then it just uses the current dir?

grayside:
> So you are saying that before this PR the root for the sync dir was whatever dir you were in when you ran the command, and not the directory that the outrigger.yml file is in?

Correct, if you look at the code, it used `os.Getwd()` pretty ubiquitously.

> Now with this PR it will only work if there is an outrigger.yml and use that file's directory, but we also need to account for the current working directory if that project config file doesn't exist?

Yes, that's what my latest comment discussed. I over-corrected. We wanted to be able to use this functionality outside a project context. That seems like an edge case (similar to us using `docker run` instead of `docker-compose run`), but still probably worth supporting for quick experiments.

> Also, with the issue you linked, recursive project config discovery, that seems kind of problematic. Maybe the project config file need to specify the sync dir, and if there is no project config in the recursive heirarchy that specifies a sync dir then it just uses the current dir?

The idea is that if an outrigger.yml is found, the rig project sync should be rooted in that directory. If not, it should be the current working directory. It makes sense to me that we further enhance the outrigger.yml unison config to specify a relative path to sync from below the project config structure. Currently all rig project scripts execute in the directory of the outrigger.yml file, we assume it is always at the repo root.

febbraro:
But it is possibly to have a "global" outrigger.yml, say in your $HOME dir, and it is possible without a project config (for various use cases) that it will share your $HOME dir via unison.  Should we additionally add a flag that says use $CWD?

grayside:
I think it makes sense to add a flag to designate the localhost side of the sync as a path. So `--sync-from=.` or `--sync-from=$PWD` would end up as the current directory, but other options are available, such as if a particular project is structured such that only the contents of a subdirectory should be synced.

In priority order, this gives us: 

1. --sync-from flag
2. outrigger.yml
3. Current working directory

As long as we're talking about these options, it's worth noting that we currently have an argument that causes problems for other commands to interact with the sync container, so we should be careful to keep those use cases in mind. #85 for discussion on that.

febbraro:
Would `--sync-path` or `--sync-root` work better?

grayside:
Re-introduced the CWD fallback, added --sync-path override, added verbose logging of what the working directory is for sync execution, followed that through to do similar in project scripts and tweaked a couple log messages for easier reading flow.

cli/commands/project_sync.go
```diff
@@ -142,30 +168,32 @@ func (cmd *ProjectSync) StartUnisonSync(ctx *cli.Context, volumeName string, con
 		}
 	}
 	cmd.out.Verbose.Printf("Unison Args: %s", strings.Join(unisonArgs[:], " "))
-	if err = exec.Command("unison", unisonArgs...).Start(); err != nil {
+	command := exec.Command("unison", unisonArgs...)
+	// While the logFile is an absolute path to the workingDirectory, the Unison
+	// process is still watching the "current directory" and so needs the working
+	// directory to be explicitly set.
+	command.Dir = workingDir
+	cmd.out.Verbose.Printf("Script execution - Working Directory: %s", workingDir)
```
febbraro: Log message appears wrong. This should be `Sync` or something besides `Script execution`

grayside:
* [x] Log message appears wrong. Script execution -> Sync execution

cli/commands/project_sync.go
```diff
@@ -267,10 +295,10 @@ func (cmd *ProjectSync) WaitForUnisonContainer(containerName string, timeoutSeco
 
 // The local unison process is finished initializing when the log file exists
 // and has stopped growing in size
-func (cmd *ProjectSync) WaitForSyncInit(logFile string, timeoutSeconds int, syncWaitSeconds int) {
+func (cmd *ProjectSync) WaitForSyncInit(logFile string, workingDir string, timeoutSeconds int, syncWaitSeconds int) {
 	cmd.out.Info.Print("Waiting for initial sync detection")
 
-	var tempFile = fmt.Sprintf(".%s.tmp", logFile)
+	var tempFile = filepath.Join(workingDir, fmt.Sprintf(".test-sync-start.tmp"))
```
febbraro: Don;t need a Sprintf here is the log file name is not patterned from a variable. `%s` or whatevs.

cli/commands/project_sync.go
```diff
@@ -65,33 +72,50 @@ func (cmd *ProjectSync) Commands() []cli.Command {
...
 
-// Start the unison sync process
+// Start the Unison sync process.
 func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
 	cmd.Config = NewProjectConfig()
 	cmd.out.Verbose.Printf("Loaded project configuration from %s", cmd.Config.Path)
-	volumeName := cmd.GetVolumeName(ctx, cmd.Config)
+	// Determine the working directory for CWD-sensitive operations.
+	var workingDir string
+	if syncPath, err := cmd.DeriveLocalSyncPath(cmd.Config, ctx.String("sync-path")); err != nil {
```
febbraro: could you not use `workingDir` here as it is already defined as a var and avoid the else all together?

cli/commands/project_sync.go
febbraro: I think what I meant to say is, why are you not using `logFile`

cli/commands/project_sync.go
```diff
@@ -323,3 +351,36 @@ func (cmd *ProjectSync) GetUnisonMinorVersion() string {
 	segments := v.Segments()
 	return fmt.Sprintf("%d.%d", segments[0], segments[1])
 }
+
+// Remove the Unison logfile.
+func (cmd *ProjectSync) RemoveLogFile(logFile string, workingDir string) error {
+	command := exec.Command("rm", "-f", logFile)
+	command.Dir = workingDir
```
febbraro: If `logFile` is a variable with `workingDir` already joined, you dont need to pass in `workingDir` and set it the `command.Dir`

cli/commands/project_sync.go
febbraro: Also, `logFile` should already be a full absolute path, so I'm thinking that the `filepath.Join` is not needed.

grayside: This logic predates the PR. I think logFile is used by unison, and we're checking that tempFile is synced. I wonder what happens if a project ignores *.tmp files... might need a parameter or less fraught ext.

cli/commands/project_sync.go
grayside: I declare the var so it's in function instead of block scope, but the := operator is caught by the compiler as an overwrite. I could nest the whole function inside the if with reversed err check.

grayside:
* [x] Improved code flow around workingDir by nesting login inside if blocks.

cli/commands/project_sync.go
grayside: My reasoning here is that if it is worth having this broken out as a separate function, and an exported one at that, it makes sense to not make assumptions about the logFile path. In fact, you could argue that this should be checking the logFile path exists.

I don't have strong feelings about folding it back in or helping it to stand alone better. Do you have a preference?

febbraro: I think we can keep it, I did not realize you intended to make that funciton for public consumption.  If so, then don't call it `RemoveLogFile` just call it `RemoveFile` and rename the variable and then things make more sense as a general purpose function.

cli/commands/project_sync.go
```diff
@@ -38,14 +41,14 @@ func (cmd *ProjectSync) Commands() []cli.Command {
 	start := cli.Command{
 		Name:        "sync:start",
 		Aliases:     []string{"sync"},
-		Usage:       "Start a unison sync on local project directory. Optionally provide a volume name.",
+		Usage:       "Start a Unison sync on local project directory. Optionally provide a volume name.",
```
grayside: Did some renaming so Unison is capitalized in messages.

cli/commands/project_sync.go
```diff
@@ -122,10 +147,14 @@ func (cmd *ProjectSync) StartUnisonSync(ctx *cli.Context, volumeName string, con
 
 	cmd.out.Info.Println("Initializing sync")
 
-	// Remove the log file, the existence of the log file will mean that sync is up and running
+	// Determine the location of the local Unison log file.
 	var logFile = fmt.Sprintf("%s.log", volumeName)
-	exec.Command("rm", "-f", logFile).Run()
```
grayside: Created util/filesystem.go for these basic kinds of operations.

cli/commands/project_sync.go
```diff
@@ -267,49 +292,65 @@ func (cmd *ProjectSync) WaitForUnisonContainer(containerName string, timeoutSeco
...
-	var timeoutLoopSleep = time.Duration(100) * time.Millisecond
-	// * 10 here because we loop once every 100 ms and we want to get to seconds
-	var timeoutLoops = timeoutSeconds * 10
-
-	// Create a temp file to cause a sync action
-	exec.Command("touch", tempFile).Run()
+	// The use of os.Stat below is not subject to our working directory configuration,
+	// so to ensure we can stat the log file we convert it to an absolute path.
+	if logFilePath, err := util.AbsJoin(workingDir, logFile); err != nil {
+		cmd.out.Info.Print(err.Error())
```
grayside: I've attempted to fix the inconsistency between absolute paths vs. working direct & relative paths throughout this file, but os.Stat is an edge case that needs support.

cli/commands/project_sync.go
```diff
@@ -267,49 +292,65 @@ func (cmd *ProjectSync) WaitForUnisonContainer(containerName string, timeoutSeco
...
-			var statSleep = time.Duration(syncWaitSeconds) * time.Second
-			// Initialize at -2 to force at least one loop
-			var lastSize = int64(-2)
-			for lastSize != statInfo.Size() {
+		cmd.out.Verbose.Printf("Creating temporary file so we can watch for Unison initialization: %s", tempFile)
+
+		var timeoutLoopSleep = time.Duration(100) * time.Millisecond
+		// * 10 here because we loop once every 100 ms and we want to get to seconds
+		var timeoutLoops = timeoutSeconds * 10
+		var statSleep = time.Duration(syncWaitSeconds) * time.Second
```
grayside: Moved statSleep up because we were previously looping over an otherwise unchanging operation.

cli/commands/project_sync.go
```diff
@@ -267,49 +292,65 @@ func (cmd *ProjectSync) WaitForUnisonContainer(containerName string, timeoutSeco
...
-
-	// Create a temp file to cause a sync action
-	exec.Command("touch", tempFile).Run()
+	// The use of os.Stat below is not subject to our working directory configuration,
+	// so to ensure we can stat the log file we convert it to an absolute path.
+	if logFilePath, err := util.AbsJoin(workingDir, logFile); err != nil {
+		cmd.out.Info.Print(err.Error())
+	} else {
+		// Create a temp file to cause a sync action
+		var tempFile = ".rig-check-sync-start"
```
grayside: Renamed this further to remove .tmp suffix and introduce rig in the name to clarify what it is and avoid potential sync ignores around not syncing tempfiles.

cli/util/filesystem.go
```diff
@@ -0,0 +1,53 @@
+package util
```
grayside: All new code in this file.

febbraro:
So, it will choose either the `sync-path` provided, or the location of the config file, or the current working directory.  Do we think that it makes sense to have a generally configurable path in the sync section of the config file? or do you think that is just generally captured under the `--sync=path` option and you'd encod ethat directly into your project scripts?  

I will do a bit of a deeper review tonight, but i'm thinking this thing is just about good to merge, thanks for all the hard work on it.

grayside:
@febbraro my thinking is that projects will always sync relative to the outrigger.yml, because in that way if you connect via the CLI container all the pieces are the same, which I find tremendously valuable for general intuition. I could see the value of a sync-path option in outrigger.yml, but given the ignorenot modifier for Unison we theoretically support it already. https://www.cis.upenn.edu/~bcpierce/unison/download/releases/stable/unison-manual.html#ignore

grayside:
Travis failed the branch on a partial refactoring I had in the util directory, to take the easy way out I added those changes to this PR.

I've broken up util.go into user_input.go and docker.go and folded image.go into docker.go. When there's more than one file I think semantically naming and categorizing everything makes it easier to find.

grayside:
This one wins the award for hardest easy PR I've landed in rig.

Issue #88, opened 2017-10-17
[Project Config] Verbose logging not available in Commands() calls


grayside:
The logger is initialized on app.Before, which happens after commands are loaded and before the process of executing the selected command begins. As a result, cmd.Commands() is called for each command file before the verbose mode is activated. In fact, if any Command() function attempts to instantiate the logger it will keep logging non-verbose for the rest of the command execution because of the logger's singleton pattern.

After some consideration I do not see a way to change this. Commands() needs to be called as part of the cli bootstrap. Anything called from Commands() needs to carefully not use the Logger class.

This seems like something we need to settle and put into a nascent CONTRIBUTING.md

grayside:
#89 implements a bandaid for this by re-initializing the logger on command.Before. In that way whatever happens in Commands() is reset, and we just have a blanket policy of no Verbose logging in Commands().

PR #89, opened 2017-10-17
Recursively find outrigger.yml

grayside:
This adjusts the way project config is loaded by default. It will first check the current working directory... then it will traverse up through the directory structure, much like git looking for a .git directory.

As part of this, I wanted to ensure verbose logging of the location of the loaded file, which led me down the path of investigating #88. I'm not quite sure if I'm not seeing things that this extra initialization does something. Also interesting, the verbose log in cmd.Before that I've added to project_sync.go and project.go is being called twice.

The screenshot illustrates what this does even better:
recursive-config-discovery
<sup>In the screenshot, you see the config is loaded from a parent directory of the PWD and executing as though in that parent directory.</sup>

Fix for #87, partial fix for #88.
I envision this being pushed through before the refactoring to get #86 wrapped up.

### DIFF
--- a/cli/commands/project.go
+++ b/cli/commands/project.go
@@ -42,6 +42,16 @@ func (cmd *Project) Commands() []cli.Command {
 	return []cli.Command{command}
 }
 
+// Run before all commands to setup core services
+func (cmd *Project) Before(c *cli.Context) error {
+	if err := cmd.BaseCommand.Before(c); err != nil {
+		return err
+	}
+	cmd.out.Verbose.Printf("Loaded project configuration from %s", cmd.Config.Path)
+
+	return nil
+}
+
 // Processes script configuration into formal subcommands.
 func (cmd *Project) GetScriptsAsSubcommands(otherSubcommands []cli.Command) []cli.Command {
 
--- a/cli/commands/project.go
+++ b/cli/commands/project.go
@@ -42,19 +42,8 @@ func (cmd *Project) Commands() []cli.Command {
 	return []cli.Command{command}
 }
 
-// Run before all commands to setup core services
-func (cmd *Project) Before(c *cli.Context) error {
-	if err := cmd.BaseCommand.Before(c); err != nil {
-		return err
-	}
-	cmd.out.Verbose.Printf("Loaded project configuration from %s", cmd.Config.Path)
-
-	return nil
-}
-
 // Processes script configuration into formal subcommands.
 func (cmd *Project) GetScriptsAsSubcommands(otherSubcommands []cli.Command) []cli.Command {
-
 	cmd.Config.ValidateProjectScripts(otherSubcommands)
 
 	if cmd.Config.Scripts == nil {
@@ -77,7 +66,6 @@ func (cmd *Project) GetScriptsAsSubcommands(otherSubcommands []cli.Command) []cl
 			if len(script.Alias) > 0 {
 				command.Aliases = []string{script.Alias}
 			}
-
 			command.Description = command.Description + cmd.ScriptRunHelp(script)
 
 			commands = append(commands, command)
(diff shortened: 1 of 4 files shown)


cli/commands/project.go
```diff
@@ -42,6 +42,16 @@ func (cmd *Project) Commands() []cli.Command {
 	return []cli.Command{command}
 }
 
+// Run before all commands to setup core services
+func (cmd *Project) Before(c *cli.Context) error {
+	if err := cmd.BaseCommand.Before(c); err != nil {
+		return err
+	}
+	cmd.out.Verbose.Printf("Loaded project configuration from %s", cmd.Config.Path)
```
febbraro: Why not just print this out at the start of Run()?

cli/commands/project_sync.go
```diff
@@ -71,18 +72,28 @@ func (cmd *ProjectSync) Commands() []cli.Command {
 	return []cli.Command{start, stop}
 }
 
+// Run before all commands to setup core services
+func (cmd *ProjectSync) Before(c *cli.Context) error {
+	if err := cmd.BaseCommand.Before(c); err != nil {
+		return err
+	}
+	cmd.Config = NewProjectConfig()
```
febbraro: Why not just do this at the start of `RunStart()` and `RunStop()`, we can make an `InitProjectConfig()`, then we dont have to deal with the BaseCommand shuffling.

cli/commands/project.go
grayside: The idea is that for a given file we want every command to load and log configuration the same way. This seemed slightly DRY-er. I briefly explored the idea of having the project set a flag to enable config and do this just in BaseCommand but pulled back on the basis of scope.

cli/commands/project_sync.go
grayside: Sure, we can do an InitProjectConfig() approach.

cli/commands/project.go
febbraro: For the project command, the config is loaded in Commands() so the config is loaded already and every script passes through Run() before it gets executed, so I would think either moving the log message to Run() so we dont need the BaseCommand delegation or move the log to Commands() after it is loaded to the load/log are in the same place (I realize in Commands() you dont have access to the verbose flag, so think moving to Run() is the right call)

grayside:
Incorporated feedback around removing shadowed Before functions in favor of just logging/loading config at the top of Action callback.

cli/commands/project_config.go
```diff
@@ -35,33 +36,48 @@ type ProjectConfig struct {
...
-			break
 		}
 	}
 
 	return readyConfig
 }
 
-// Create a new ProjectConfig from the specified file
+// Traverse directory structure looking for an outrigger project config file.
+func FindProjectConfigFilePath() (string, error) {
```
tekante: If I understand the code correctly, this could ascend up and out of the project directory and pick up an outrigger file that may be lingering around in a directory above where the project is checked out. Seems like an unlikely edge case (should only happen if someone is trying to use a rig command in a project that isn't set up for it). I'm not sure an effective way to prevent against it (maybe don't ascend above any directory with a .git directory in it) but should we document somewhere that this is possible?

In thinking through whether we could make a recommendation to have an outrigger.yml file at the user directory level or their "Project" directory to catch when this happens and alert people I can't think of a mechanism where that wouldn't just cause confusion as there isn't a way to put in a global echo command I can think of.

Perhaps just a verbose logging of what outrigger file is being used is the path isn't os.Getwd()?

tekante: I see that the logging is down below in the unison sync process setup and above in the project scripts command. I had missed it before so it looks like this is already handled.

Issue #68, opened 2017-08-30
Migrate from godep to dep


febbraro:
`dep` is quickly becoming the package manager of choice for the go project (with goals to integrate it into the core product).

Transition from godep to dep for managing our external dependencies.

grayside:
I see from the [contributors](https://github.com/golang/dep/graphs/contributors) that this is part of the Golang dependency management initiative that Sam Boyer is a leader of.

PR #94, opened 2017-10-20
Upgraded to Go 1.9 and dep

febbraro:
Move away from Godep to dep (thanks to @sdboyer)

This is for #68


cli/Gopkg.lock
```diff
@@ -0,0 +1,56 @@
...
+  name = "golang.org/x/sys"
+  packages = ["unix"]
+  revision = "eb2c74142fd19a79b3f237334c7384d5167b1b46"
+
+[[projects]]
+  name = "gopkg.in/yaml.v2"
+  packages = ["."]
+  revision = "cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b"
+
+[solve-meta]
```
grayside: Not sure where it belongs, but depending readme recommends ci should use a fixed version.

grayside: `dep` readme, not depending

febbraro:
@grayside Updated based on feedback. Multiple dependencies did not have explicit versions, so I pinned to a revision

sdboyer:
hola friends! couple quick notes on this:

1. you'll probably have an easier time if you use release versions of dep instead of tip.
2. if there's ever a possibility that something else will depend on rig, you'll likely want to place the `Gopkg.{toml,lock},vendor` at the repository root.
3. in general, you needn't specify a `revision` in `Gopkg.toml` - instead, trust `Gopkg.lock` to pin things appropriately. if there's no tags to consume, prefer a `branch`. then just make sure to always use `dep ensure -update <deps...>` instead of `dep ensure -update` (globally), and those pinned versions won't change.

febbraro:
Thanks @sdboyer! Appreciate the feedback

febbraro:
Moving all this work to #95

PR #95, opened 2017-10-21
GoReleaser support

febbraro:
This PR is built on the back of #94, so we can either push that one first and then this, or abandon that and do the work of both in this PR.

This adds GoReleaser support which does the following
* Compiles for our 3 targets (mac, linux, windows)
* Creates archives of those binaries
* Publishes them and a changelog to GitHub Releases
* Publishes a new version to Homebrew (need to figure out where it puts the downloadable)
* Creates a .deb and an .rpm for installing on Debian/Ubuntu and Fedora/RHEL/CentOS


febbraro:
Looks like it pulls the homebrew release binary directly from GitHub

build.sh
```diff
@@ -2,12 +2,12 @@
 #
 # To use this you will need to: go get github.com/mitchellh/gox
 
-go get github.com/tools/godep
-go get github.com/mitchellh/gox
+go get -u github.com/golang/dep/...
```
grayside: What do the 3 dots indicate? Shouldn't we be specifying a specific version?

cli/Gopkg.toml
```diff
@@ -0,0 +1,50 @@
...
+#  version = "2.4.0"
+
+
+[[constraint]]
+  name = "github.com/bitly/go-simplejson"
+  version = "0.5.0"
+
+[[constraint]]
+  name = "github.com/fatih/color"
+  version = "1.5.0"
```
grayside: Some of these versions are specific, is that on purpose vs. branches?

PR #97, opened 2017-10-25
Initial test coverage in util package.

grayside:


grayside:
Despite my reading, it seems that you can only have one TestMain in a given test run. Not sure if that means across the entire codebase or an individual sub-package under test or what. Continuing to investigate.

grayside:
Okay, back to ready, and green.

PR #98, opened 2017-10-26
Project Sync transitioned to new success/error handling

febbraro:
This was the big daddy, but I think I got it.  Required a lil bit of refactoring to make it all work.

### DIFF
--- a/cli/commands/project_sync.go
+++ b/cli/commands/project_sync.go
@@ -96,7 +96,6 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
 		cmd.out.Verbose.Printf("Loaded project configuration from %s", cmd.Config.Path)
 	}
 
-
 	// Determine the working directory for CWD-sensitive operations.
 	var workingDir, err = cmd.DeriveLocalSyncPath(cmd.Config, ctx.String("dir"))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cli/commands/project_sync.go
```diff
@@ -97,56 +97,60 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
...
-		case "linux":
-			cmd.out.Verbose.Printf("Setting up local volume: %s", volumeName)
-			cmd.SetupBindVolume(volumeName, workingDir)
-		default:
-			cmd.out.Verbose.Printf("Starting sync with volume: %s", volumeName)
-			cmd.StartUnisonSync(ctx, volumeName, cmd.Config, workingDir)
-		}
-	} else {
-		cmd.out.Error.Fatal(err)
+	var workingDir, err = cmd.DeriveLocalSyncPath(cmd.Config, ctx.String("dir"))
```
grayside: Why does this work without the `:=`? When I've tried things like this before, I'd have err yelling at me.

cli/commands/project_sync.go
```diff
@@ -97,56 +97,60 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
...
+	if err != nil {
+		return cmd.Error(err.Error(), "SYNC-PATH-ERROR", 12)
```
grayside: Hm, maybe we should define cmd.EXTERNAL_ERROR_CODE and so on for more clarity about the numbers? I realize that's outside immediate scope but reading this out of the context of the other PR this immediately jumps out at me as "why is this a number 12?"

cli/commands/project_sync.go
febbraro: := is just a shorthand for not needing to `var` it. The var and = I believe is redefining it.

cli/commands/project_sync.go
febbraro: I think some constants for standard error codes are probably a good idea. We can tackle that in another PR

PR #99, opened 2017-10-26
Linux compatibility

febbraro:


cli/commands/dashboard.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	dashboardContainerName string = "outrigger-dashboard"
-	dashboardImageName     string = "outrigger/dashboard:latest"
```
grayside: I made these lower camel case to keep them as private constants. Do we have a use case to export them?

cli/commands/dashboard.go
```diff
@@ -30,7 +30,7 @@ func (cmd *Dashboard) Commands() []cli.Command {
 }
 
 func (cmd *Dashboard) Run(ctx *cli.Context) error {
-	if cmd.machine.IsRunning() {
+	if cmd.machine.IsRunning() || runtime.GOOS == "linux" {
```
grayside: Nice catch.

cli/commands/data_backup.go
```diff
@@ -37,8 +38,12 @@ func (cmd *DataBackup) Commands() []cli.Command {
 }
 
 func (cmd *DataBackup) Run(c *cli.Context) error {
+	if runtime.GOOS == "linux" {
```
grayside: Given the need to add runtime and check on Linux everywhere, I wonder if we shouldn't add a util/feature_detection.go now and define util.CanRunNative() as a check for linux.

cli/commands/data_restore.go
```diff
@@ -37,8 +38,12 @@ func (cmd *DataRestore) Commands() []cli.Command {
 }
 
 func (cmd *DataRestore) Run(c *cli.Context) error {
+	if runtime.GOOS == "linux" {
+		return cmd.Success("Data Restore is not needed on Linux, please unarchive any data directly")
```
grayside: This makes sense to me, but if we ever expanded on the data backup to include volume wrangling this would change.

cli/commands/dashboard.go
febbraro: Did not mean to make them public (by convention), was just looking at some approaches. i can revert that.

cli/commands/data_backup.go
febbraro: `util.NativeDockerSupport()` ?

grayside: In anticipation of other things to detect, `util.SupportsNativeDocker()`.

`util.SupportsFileSync()` would currently exclude windows without bash right?

PR #100, opened 2017-10-26
Removed watch command

febbraro:
use unison instead


tekante:
Is this slated for a 2.0 version? If not seems like something we might consider having issue a deprecation message about not being needing and projects should switch to unison for anyone who may have a set of scripts trying to call this. At minimum noting it as a significant adjustment in the changelog and release announcement.

febbraro:
Yeah this next release will be 2.0 and we have had a deprecation warning in the watch command since we released the unison support.

PR #101, opened 2017-10-27
Repo reorg

febbraro:
Repo is now reorg'd to not have all the code in a subfolder.


.travis.yml
```diff
@@ -12,11 +12,9 @@ matrix:
   allow_failures:
     - go: tip
 
-before_install: cd cli
-
 script:
-  - "../scripts/test-go-fmt.sh"
-  - "go run main.go"
+  - "scripts/test-go-fmt.sh"
+  - "go run cmd/main.go"
```
grayside: Should we just drop main.go in the root? Not sure what the cmd subdirectory adds in that case. Other than that minor thought, this looks grand.

febbraro: I looked for inspiration form other go based CLI projects and this was a convention they used pretty regularly.  Also, GoReleaser defaults to the repo root for main.go, but they show an example of overriding it and that example uses `cmd/main/go`

I think I just didnt; want any go files in the root where I expect to see config and other things.

grayside: Done and done.

Issue #92, opened 2017-10-20
Investigate use of gometalinter


grayside:
Mentioned in https://carlosbecker.com/posts/golang-cli-apps/, gometalinter aggregates and normalizes the output of various linting tools. Adding it to the project for local dev & CI seems wise, though it is sure to start with a number of violations.

PR #102, opened 2017-10-27
Add gometalinter

febbraro:
added linting and addressed all feedback (and ignored 3)

Addresses #92


commands/doctor.go
```diff
@@ -137,7 +140,7 @@ func (cmd *Doctor) Run(c *cli.Context) error {
 		output, err := exec.Command("docker-machine", "ssh", cmd.machine.Name, "df -h 2> /dev/null | grep /dev/sda1 | head -1 | awk '{print $5}' | sed 's/%//'").Output()
 		if err == nil {
 			dataUsage := strings.TrimSpace(string(output))
-			if i, err := strconv.Atoi(dataUsage); err == nil {
+			if i, e := strconv.Atoi(dataUsage); e == nil {
```
grayside: Why e?

util/feature_detection.go
```diff
@@ -4,7 +4,25 @@ import (
...
+	VirtualBox = "virtualbox"
+)
+
+// SupportsNativeDocker determines if the runtime OS support docker natively,
+// versus needing to run docker in a virtual machine
 func SupportsNativeDocker() bool {
 	return runtime.GOOS == "linux"
 }
+
+// IsMac detects if we are running on the darwin platform
```
grayside: Shouldn't we have IsLinux?
Making the other checks more feature driven seems a fine followup

grayside:
Did I miss travis config?

commands/doctor.go
febbraro: The linter was giving warnings about the second `err` shadowing the one from it's parent scope.

util/feature_detection.go
febbraro: Are you sugesting an IsLinux in addition to SupportsNativeDocker()?  Most of the linux checks are for a feature (native docker support) where as some of the mac checks were for things like which route command to use or this process to kill to restart dns resolution. So there is a bit of the mix in platform name and feature detection.  This is the quicked line to remove linter warnings and also make the code more readable (and centralized if for some crazy reason the string "darwin" goes away.

febbraro:
I didn;t add travis configuration yet. I was not sure if we wanted to fail the build on linter warnings.  Maybe we should?  I had to exempt 3 functions from linting, one due to some nested error conditions that was not ready mentally to unpack and 2 for cyclomatic complexity.  They should all be reworked but I wanted to get the superficial warnings covered and then take an separate PR to attack the complexity issues in Doctor and ProjectConfig validation.

febbraro:
Added travis integration for the linter

util/feature_detection.go
grayside: My thinking was for API completeness if we have checks for the OS's we may as well have one for each that we support.

Issue #58, opened 2017-07-24
rig project doctor


grayside:
Should we have a rig project doctor?

I've gone back and forth on this a bit: Should it be a generated script? Should it be in rig? A separate tool? Extensible? Runnable from Jenkins? Lots of questions.

With this functionality, we may be able to automate checks for a lot more of the problems projects run into and fasttrack diagnoses.

Here's the specification of what I had in mind:

```yaml
doctor:
  conditions:
    -
      name: Valid Docker Compose Configuration
      diagnosis: Your docker-compose files do not provide a valid configuration.
      prescription: Evaluate the errors and revise the configuration files as needed.
      # If a check fails on a fatal severity issue, stop the process.
      severity: error
      check:
        - validate-docker-compose.sh
    -
      name: Scaffolded Files Exist
      diagnosis: Some of the generated files for your project have gone missing. This can lead to unpredictable results.
      prescription: Replace missing files from git history or a rerun of the generator. If any file is not needed, please delete from this check.
     severity: warning
     # Checks should be run sequentially and the whole step failed if any one fails.
     check:
       - ls src > /dev/null
       - ls package.json > /dev/null
```

### Unison Sync Crash

* Diagnosis: Your unison sync crashed and will not work without further action.
* Prescription: Stop your project containers, run `docker container rm -f projectname-sync` and `docker volume rm -f projectname-sync`
* Severity: Fatal

### Containers Down

* Diagnosis: No containers associated with the projectname are currently operating.
* Prescription: Start your project containers before relying on their services.
* Severity: Warning

grayside:
I 'reopened' the idea of doing this when I got most of the way through a change to rig to allow rig project sync:start to be run with a flag to first wipe the existing sync container, before realizing that was black magic for the team to learn, and something to instruct them on the directly docker way to handle the problem would be better for learning purposes.

PR #103, opened 2017-10-29
Removed volume name as an arg to sync:start

febbraro:
An example of what would be needed for #85


grayside:
This makes sense to me. It means the only edge case on finding the sync volume for a project is whether the project uses the `--dir` and does not configure a volume name. That seems like something for #58/documentation to incorporate as best practice guidance.

#85 includes the more outrageous proposal of taking the idea of #6 to have a `rig data` or `rig fs` command, and include ad hoc unison sync volume management as a sub-component there and require rig project sync to be explicitly configured. I think that can wait for use cases/v3.

PR #111, opened 2017-10-31
Display executed commands via Verbose Logging

### LINKED ISSUE #5: Output all executed commands when --verbose and explain

grayside:
Finally getting back to the learnification.

This is still in-progress but wanted to get an approach confirmation before continuing.

3__thanks_for_flying_vim__zsh_

Fixes #5

### DIFF
--- a/util/shell_exec.go
+++ b/util/shell_exec.go
@@ -75,7 +75,6 @@ func (x Executor) Execute(forceOutput bool) error {
 		x.cmd.Stdout = ioutil.Discard
 	}
 
-	x.Log("Executing")
 	color.Set(color.FgCyan)
 	err := x.Run()
 	color.Unset()
@@ -84,7 +83,20 @@ func (x Executor) Execute(forceOutput bool) error {
 
 // Run runs a command via exec.Run() without modification or output of the underlying command.
 func (x Executor) Run() error {
+	x.Log("Executing")
 	return x.cmd.Run()
+}
+
+// Output runs a command via exec.Output() without modification or output of the underlying command.
+func (x Executor) Output() ([]byte, error) {
+	x.Log("Executing")
+	return x.cmd.Output()
+}
+
+// CombinedOutput runs a command via exec.CombinedOutput() without modification or output of the underlying command.
+func (x Executor) CombinedOutput() ([]byte, error) {
+	x.Log("Executing")
+	return x.cmd.CombinedOutput()
 }
 
 // Log verbosely logs the command.
--- a/util/shell_exec.go
+++ b/util/shell_exec.go
@@ -18,15 +18,16 @@ type Executor struct {
 }
 
 // StreamCommand sets up the output streams (and colors) to stream command output if verbose is configured
-func StreamCommand(cmd *exec.Cmd) error {
-	return Executor{cmd}.Execute(false)
+func StreamCommand(path string, arg ...string) error {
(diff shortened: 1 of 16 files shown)


febbraro:
Maybe we should use `util.Run` for everything instead of `exec.Command` and then we can just create the command to run?  like maybe util.Run returns an `exec.Cmd` so you can still choose to `Run` or `CombineOutput` or whatever?  I realize we dont always handle Cmds in a consistent way, but seems more clear/simple than wrapping every `exec.Command` call

grayside:
The problem there is if we output the log when we create the command, that could be out of sequence when we actually execute it. Certainly wrapping the exec.Command factory is part of the strategy in #97 so I don't object on that point.

Would it make sense to extend exec.Cmd into our own struct and shadow Run et al with the verbose logging?

febbraro:
Shadowing might be good, sometimes we don't call `Run()` directly though, but call `CombinedOutput()` or `Output()`, I wonder if shadowing `Run()` would get our method called, or the base struct?

grayside:
To do this at exec time, I assume we need to shadow each unless the API internals use a common exec call we can shadow.

grayside:
Moved command output to yellow for better visibility. Shifted to more of an object wrapper model. Stopped implementation for approach review.

util/shell_exec.go
```diff
@@ -69,3 +64,49 @@ func PassthruCommand(cmd *exec.Cmd) (exitCode int) {
...
+	x.Log("Executing")
+	color.Set(color.FgCyan)
+	err := x.Run()
+	color.Unset()
+	return err
+}
+
+// Run runs a command via exec.Run() without modification or output of the underlying command.
+func (x Executor) Run() error {
+	return x.cmd.Run()
```
febbraro: Do we want to log all commands that are run? Or only commands that force/stream?

febbraro:
I like how this is looking a lot better.  The code that needs to run commands look really no different and is still pretty easy to read, and the Executor looks pretty simple.

util/shell_exec.go
```diff
@@ -2,38 +2,33 @@ package util
...
 const defaultFailedCode = 1
 
+type Executor struct {
+	cmd *exec.Cmd
+}
+
 // StreamCommand sets up the output streams (and colors) to stream command output if verbose is configured
 func StreamCommand(cmd *exec.Cmd) error {
-	return RunCommand(cmd, false)
+	return Executor{cmd}.Execute(false)
```
febbraro: You know, we could make this and `ForceStreamCommand` even easier to read on the caller side if we just take the command and args here instead of the `exec.Cmd` and call `exec.Command` just like the `Executor.Execute`

util/shell_exec.go
grayside: Probably all commands, but I wanted to lay out the initial API for it more flexibly so the refactoring can go either way.

grayside:
Too much?

3__root_bbefb61ec982__var_www_northwell-d8__zsh_

grayside:
Note that in the example above, we see that our use of machine.GetData() could maybe do some static caching. Likely a nice little performance optimization.

PR #112, opened 2017-10-31
Linux DNS install

febbraro:


commands/dns.go
```diff
@@ -77,40 +77,77 @@ func (cmd *DNS) RemoveHostFilter(ipAddr string) {
...
+		util.StreamCommand(exec.Command("bash", "-c", fmt.Sprintf("echo 'server=/vm/%s' | sudo tee /etc/NetworkManager/dnsmasq.d/dnsdock.conf", bridgeIP)))
+
+		// Restart NetworkManager if it is running
+		if err := exec.Command("systemctl", "is-active", "NetworkManager").Run(); err != nil {
+			util.StreamCommand(exec.Command("sudo", "systemctl", "restart", "NetworkManager"))
 		}
 	}
+
+	// Is libnss-resolver in use
+	if _, err := os.Stat("/etc/resolver"); err == nil {
```
grayside: To be clear, we will set up *.vm support via both these mechanisms if both happen to be present? I believe we discussed that being safe and the approach, but wanted to confirm.

test/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+* `GOARCH=amd64 GOOS=linux go build -o build/linux/rig cmd/main.go`
+
+
+## Run the images (and Docker in Docker)
+
+* Start Docker in Docker
+    * `docker run --privileged -it --name dind -d docker:dind`
+* Start the container for the distro you want, mounting a linux targeted `rig` into it and linking it to the Docker in Docker image.
+    * `docker run -it -v $PWD/build/linux/rig:/usr/bin/rig --link dind:docker test-centos bash`
```
grayside: Does travis support dind?

febbraro: I believe it does

commands/dns.go
```diff
@@ -41,20 +40,56 @@ func (cmd *DNS) Commands() []cli.Command {
...
+	if util.IsMac() {
+		cmd.configureMacRoutes(machine)
+	} else if util.IsWindows() {
+		cmd.configureWindowsRoutes(machine)
+	}
+}
+
+// ConfigureMac configures DNS resolution and network routing
+func (cmd *DNS) configureMacRoutes(machine Machine) {
+	cmd.out.Verbose.Print("Configuring network routing for macOS")
```
grayside: Is this log valuable, as it directly follows this info log
> Setting up local networking (may require your admin password)

commands/dns.go
```diff
@@ -41,20 +40,56 @@ func (cmd *DNS) Commands() []cli.Command {
...
+	machineIP := machine.GetIP()
+
+	if machine.IsXhyve() {
+		cmd.removeHostFilter(machineIP)
+	}
+	exec.Command("sudo", "route", "-n", "delete", "-net", "172.17.0.0").Run()
+	util.StreamCommand(exec.Command("sudo", "route", "-n", "add", "172.17.0.0/16", machineIP))
```
grayside: Why is delete `Run()` and add `StreamCommand()`?

commands/dns.go
```diff
@@ -132,14 +142,84 @@ func (cmd *DNS) ConfigureDNS(machine Machine, nameservers string) {
 		"-l", "com.dnsdock.name=dnsdock",
 		"-l", "com.dnsdock.image=outrigger",
 		"--name", "dnsdock",
-		"-p", bridgeIP + ":53:53/udp",
+		"-p", fmt.Sprintf("%s:53:53/udp", bridgeIP),
```
grayside: 👍

commands/dns.go
```diff
@@ -132,14 +142,84 @@ func (cmd *DNS) ConfigureDNS(machine Machine, nameservers string) {
...
 	}
 	util.ForceStreamCommand(exec.Command("docker", args...))
+
+	// Configure the resolvers based on platform
+	var resolverReturn error
+	if util.IsMac() {
+		resolverReturn = cmd.configureMacResolver(machine)
+	} else if util.IsLinux() {
+		resolverReturn = cmd.configureLinuxResolver()
+	} else if util.IsWindows() {
```
grayside: It seems odd to have Linux in the middle, especially since it has a different signature.

util/docker.go
```diff
@@ -64,3 +64,17 @@ func ImageOlderThan(image string, elapsedSeconds float64) (bool, float64, error)
...
+// GetBridgeIP returns the IP address of the Docker bridge network gateway
+func GetBridgeIP() (string, error) {
+	output, err := exec.Command("docker", "network", "inspect", "bridge", "--format", "{{(index .IPAM.Config 0).Gateway}}").Output()
+	if err != nil {
+		return "", err
+	}
+
+	bip := strings.Trim(string(output), "\n")
+	if bip == "" {
+		bip = "172.17.0.1"
```
grayside: How do we know this bip is a safe fallback?

commands/dns.go
febbraro: Maybe not I can remove those.

commands/dns.go
febbraro: So, it has been Stream for a while, I think Stream was used to get the output to the console

commands/dns.go
febbraro: I tried to put them in order of use, meaning linux is 2nd most used.

util/docker.go
febbraro: `172.17.0.1` is the default for Docker as configured by Docker.  A user can change their default bridge IP.

febbraro:
Since all of these configs target the .vm Ltd directly, I don’t see any harm in leaving them. We could do a quick blanket remove with an uninstall command or as part of the remove command or flag on the dns command but not sure it is worth it. Maybe as a follow up at some point later?

febbraro:
Worth noting that currently we dont remove the mac settings either

grayside:
I'm not worried about uninstallation of this. I'd rather direct the time to #5 so we can more effectively help people understand the changes we are making.

PR #115, opened 2017-11-07
Add Spinner implementation for longer-running operations.

grayside:
There are a number of operations that rig command execute which can take awhile, which is exacerbated by our use of the --verbose flag to keep some of the inner workings terse. As an example, a `rig start` can go almost a minute without a single terminal output if --verbose is not used and several retries are needed to get the docker machine up.

This PR adds a new util/spinner.go mechanism which wraps a new library that provides spinner support. If rig is run in verbose mode, it will pass messages sent to the spinner to the logger, if not, it will output the messages in conjunction with a spinner lifecycle to help communicate when rig is managing a longer-running process.

3__thanks_for_flying_vim__zsh_
3__thanks_for_flying_vim__zsh_
3__thanks_for_flying_vim__sudo_
3__thanks_for_flying_vim__zsh_
3__thanks_for_flying_vim__docker_

This PR attempts to cover all the bases, with exceptions in rig project sync and machine.go::WaitForDev, which seemed just as well as follow-ups.


util/spinner.go
```diff
@@ -0,0 +1,73 @@
...
+	progress = &RigSpinner{Spins: s}
+	return progress
+}
+
+// Start restarts the spinner for a new task.
+// All functions with manage the spinner are contingent on the logger's verbosity status.
+// When verbose, we perform basic logging instead of running the spinner.
+func (s *RigSpinner) Start(message string) {
+	if Logger().IsVerbose {
+		Logger().Info.Println(message)
```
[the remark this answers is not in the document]
grayside: The problem I found without the start message is that the spinner message would sometimes lack context. Let me audit/update the messages to see if there's enough context in verbose mode withotu the start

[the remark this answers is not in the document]
febbraro: I think when the spinner is in place you should use the start message, but when in verbose mode, just do nothing.

CONTRIBUTING.md
```diff
@@ -0,0 +1,127 @@
...
+Thank you for considering contributing to the Outrigger CLI!
+
+## Quality Contributions
+
+* Make sure your branch will compile.
+* Make sure your branch passes our static analysis checks.
+* Make sure your branch conforms with go fmt standards.
+* Manually test your changes.
+
+## User Interactions
```
grayside: New CONTRIBUTING guide, which includes thoughts on logging that illustrates some of the API that's coming together.

CONTRIBUTING.md
```diff
@@ -0,0 +1,127 @@
...
+(See ./util/logging.go)
+
+Here are a few conventions:
+
+* **Starting a task that could take more than 5 seconds:**
+  * `cmd.out.Spin("Preparing the sauce")`
+* **Use the correct method to log operational results: (Pick one)**
+  * `cmd.out.Success("Sauce is Ready.")`
+  * `cmd.out.Warn("Sauce is burnt on the bottom.")`
+  * `cmd.out.Oops("Discard this sauce and try again.")`
```
grayside: Oops is temporary, there are a number of "refactoring in progress" things going on before all the specific API details are right.

commands/stop.go
```diff
@@ -66,6 +67,8 @@ func (cmd *Stop) StopOutrigger() error {
 		util.Command("sudo", "route", "-n", "delete", "-net", "172.17.42.1").Run()
 	}
 	color.Unset()
+	cmd.out.Success("Networking cleanup completed")
+	cmd.out.NoSpin()
```
febbraro: No need to `NoSpin()` here, it's handled in `BaseCommand.Success()` no?

util/logger.go
```diff
@@ -25,12 +35,17 @@ func LoggerInit(verbose bool) {
...
 		verboseWriter = os.Stdout
 	}
+
+	s, _ := spun.NewSpinner(spun.Dots)
 	logger = &RigLogger{
 		Info:      log.New(os.Stdout, color.BlueString("[INFO] "), 0),
 		Warning:   log.New(os.Stdout, color.YellowString("[WARN] "), 0),
 		Error:     log.New(os.Stderr, color.RedString("[ERROR] "), 0),
 		Verbose:   log.New(verboseWriter, "[VERBOSE] ", 0),
+		Message:   log.New(os.Stdout, " - ", 0),
```
febbraro: This doesn't appear to be used

febbraro: Sorry, i see where it is used. What is the idea behind Message?

util/shell_exec.go
```diff
@@ -114,7 +114,7 @@ func (x Executor) Start() error {
 
 // Log verbosely logs the command.
 func (x Executor) Log(tag string) {
-	color.Set(color.FgYellow)
+	color.Set(color.FgMagenta)
```
febbraro: Was this was already merged?

commands/stop.go
grayside: True.

febbraro:
This is much more along the lines of what I was thinking.

util/logger.go
grayside: It will be pulled if continuing refactor doesn't find a use for it. I wanted to stake some concepts aside from log severity for the channels.

util/shell_exec.go
grayside: I made that change here, then created a separate branch, and didn't pull it back out. This change made it into v2.

CONTRIBUTING.md
```diff
@@ -0,0 +1,127 @@
...
+* **Going to send some contextual notes to the user**:
+  1. `cmd.out.NoSpin()` if currently using the spinner.
+  2. `cmd.out.Status("Sauce exists.")`
+  3. `cmd.out.Note("This is a list item.")`
+  4. `cmd.out.Details("The ingredients of the sauce include tomato, salt, black pepper, garlic...")`
+* **Command has executed and is successful. We do not want a notification:**
+  ```
+  cmd.out.Success("Enjoy your dinner.")
+  return cmd.Success("")
```
febbraro: `cmd.out.Success()` and `cmd.Success()` is confusing. Can they be consolidated?, or can we rename one?

febbraro: We may also want a `SilentSuccess()` so we dont need to send in `""`?

grayside: yeah, or `SuccessWithNotification()`

I agree the two Success() statements is confusing. I originally had it as `Complete()`, but complete only made sense in the context of the spinner. Unfortunately this "success" is often the equivalent of an info log. I will look at the API overall and see what can be rejiggered. Now that I know I'm on the right track I can make deeper cuts.

grayside:
On the idea that this API has mostly solidified, but is going to have git conflicts if you blow on it, might we merge this and complete polishing the experience as a follow-up?

util/logger.go
```diff
@@ -42,3 +65,63 @@ func Logger() *RigLogger {
...
+	if log.IsVerbose || !log.Spinning {
+		log.Channel.Error.Println(fmt.Sprintf(format, a...))
+	} else {
+		log.Progress.Spins.SetMessage(fmt.Sprintf(format, a...))
+		log.Progress.Spins.Fail()
+	}
+}
+
+// Verbose allows Verbose logging of more advanced activities/information.
+// In practice, if the spinner can be in use verbose is a no-op.
```
tekante: Having a little trouble wrapping my brain around this comment. Is this saying that if the spinner is running verbose output won't be seen? Is that an indicator that using the --verbose flag indicates that the spinner won't run? That is my intuition but I haven't been able to perceive in the code what makes that the case.

CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+
+Here are a few conventions:
+
+* **Starting a task that could take more than 5 seconds:**
+  * `cmd.out.Spin("Preparing the sauce")`
+* **Use the correct method to log operational results: (Pick one)**
+  * `cmd.out.Info("Sauce is Ready.")`
+  * `cmd.out.Warning("Sauce is burnt on the bottom.")`
+  * `cmd.out.Error("Discard this sauce and try again.")`
+* **Going to send some contextual notes to the user**:
```
tekante: Is this example one where you always want to send a note to the user (or is the info/verbose a distinction on whether it's always versus only if --verbose is set)? If it's an indicator that the spinner needs to be disabled in those cases should we put re-enabling it after sending the message?

I don't see any indications of using NoSpin in the code so far other than in a Failure message so that makes me think perhaps I'm not interpreting how/when it needs to be used in order to use it correctly.

util/logger.go
grayside: If you run rig --verbose, you get "classic" logging. The text from Spin() is not output in verbose mode.

CONTRIBUTING.md
grayside: It could be that NoSpin needs to be added somewhere. If you do some regular logs when the Spinner is active, it suspends the spinner on line 1, outputs the log on line 2, and resumes the spinner on line 3. If you add NoSpin before logging, it does not resume on line 3.

PR #117, opened 2017-11-10
Fixing Outrigger Dashboard on Linux

febbraro:


grayside:
LOL. Because it is. Because I thought this would be outrigger-dashboard repo.

commands/doctor.go
```diff
@@ -74,6 +74,15 @@ func (cmd *Doctor) Run(c *cli.Context) error {
...
 
+	// 1.5 Ensure docker / machine is running
+	if !util.SupportsNativeDocker() {
+		if !cmd.machine.IsRunning() {
+			cmd.out.Error.Fatalf("Machine '%s' is not running. You may need to run 'rig start'", cmd.machine.Name)
+		} else {
+			cmd.out.Info.Printf("Docker Machine (%s) is running", cmd.machine.Name)
+		}
+	}
+
```
grayside: Shouldn't this come before the docker-machine env checks above?

febbraro: Well, you have the ENV setup correctly, but the machine not started.  The machine doesn't need to be started in order to test for that env configuration. I'm not against moving the check up a step though.

PR #119, opened 2017-11-11
Command logging/magenta

grayside:
Err... so yellow is the color of warning.


grayside:
This branch seems borked, let me try again...

PR #125, opened 2017-11-18
Machine data caching

febbraro:
Moved to all pointer functions, as non-pointers make a copy of the struct and the data in a pointer didn't come with it.


febbraro:
The global pattern wouldn't make a different to this error, the fact that a single intermediary function call was not a pointer caused the copy.

PR #129, opened 2017-12-13
Support RCs via goreleaser

febbraro:


grayside:
Haven't looked at this side-by-side with goreleaser.yml, is this a tweaked duplicate? Where does vendor.orig come from?

febbraro:
The diff if that I don't have a brew formula in there as there is no way to conditionally release to brew. So if we don't want the RC being our main release, we need to simply not publish to brew.

I also labeled the github release a preproduction

Issue #136, opened 2018-01-29
Update dnsdock container version


tekante:
Rig currently launches a DNSDock container at version 1.16-1 which suffers from a fatal error if a read timeout is received on the backup DNS server. See https://github.com/aacebedo/dnsdock/issues/90

Rig has restart always set so that the container comes back but during the intervening time DNS requests fail (including the original request). One way to ameliorate this is to select a different DNS server from the default 8.8.8.8 (by using the RIG_NAMESERVERS environmental variable) if connections to 8.8.8.8 are proving unstable.

Updating rig to use version 1.16-3 or better should change this to a warning as opposed to an error. It won't solve an underlying problem with a backup DNS instance being reachable but it will reduce some of the impact in terms of the service needing to fully restart.

tekante:
Fixed by #137

PR #137, opened 2018-01-29
Upgraded dnsdock to 1.16.4

febbraro:
to prevent dnsdock from exiting on a failed fallback lookup


febbraro:
I tested the failure case with 1.16.1 and 1.16.4 and dnsdock does not restart itself on 1.16.4

febbraro:
Fixes #136

PR #138, opened 2018-01-29
Add explicit privilege prompt to improve sudo UX

### LINKED ISSUE #134: New spinner obfuscates password entry requests (e.g., for sudo ops)
> I upgraded to rig 2.1.0 and just tried to stop my dev machine and thought it was borked on the second shut down step (Cleaning up local networking). I know it says "may require your admin password” but it’s not immediately clear that it’s asking for it when the cursor just changes to a key. A couple of our other colleagues also didn’t realize it was asking for a password. I thought my dev machine just wasn’t stopping. 

This may require special handling when a given command may stop to ask for sudo, unless we can preemptively escalate to admin permissions once when we know it will be needed for a given command.

grayside:
Fixes #134

This change identifies that we need to sudo and adjusts logging/spinning behaviors accordingly. Needs some more testing (e.g., with verbose) and a bit more smarts in sudo detection, hence the WIP flag. Also some UI issues around the asynchronous nature of the spinner.

I am pausing here for approach feedback.

Here's what it looks like when it operates as intended:

### Paused for Input

screen shot 2018-01-29 at 1 12 27 pm

### Success!

screen shot 2018-01-29 at 1 11 34 pm


febbraro:
So I would simply say `Administrative privileges needed` and nothing about a password, since the `Password` prompt does that already.  Look at this example, I ran `rig dns` twice. Once to get the initial prompt because I was unprivileged...

https://www.dropbox.com/s/m8twc771il6llhq/Screenshot%202018-01-29%2013.25.10.png?dl=0

And then once again, immediately, while I still had privs....

https://www.dropbox.com/s/vtx768cuhep9xk2/Screenshot%202018-01-29%2013.25.52.png?dl=0

It gave me "please enter your password" even though it just blasted right past it.  Claiming privs needed is a good plan and it gives us the newline needed for an unobscured prompt.  Claiming they are needed and blowing past it is really just a heads up and it is not confusing.

grayside:
Unfortunately here are some of the other results I'm getting, with a password needed:

screen shot 2018-01-29 at 1 42 48 pm
screen shot 2018-01-29 at 1 42 59 pm
screen shot 2018-01-29 at 1 43 25 pm

grayside:
Simplification helped tremendously. This is ready for final review.

I would appreciate if someone took it for a spin to confirm the timing issues are not still lurking.

My test has been:
```bash
./build/darwin/rig dns
```

tekante:
Looks like it's working as intended. First run through didn't need password as I had recently used sudo, second time includes the password prompt.

privilege-prompt-test-results

util/logger.go
```diff
@@ -125,3 +127,21 @@ func (log *RigLogger) Verbose(format string, a ...interface{}) {
...
+	defer func() { log.Privileged = true }()
+
+	if log.Privileged {
+		return
+	}
+
+	// This newline ensures the last status before escallation is preserved
+	// on-screen.
+	fmt.Println()
+	log.Spin("Evaluating administrative action...")
```
tekante: Just noting that things go by so quickly on my machine I never see the "Evaluating administrative action..." message.

util/slices.go
```diff
@@ -0,0 +1,13 @@
+package util
+
+// IndexOfString is a general utility function that can find the index of a value
+// present in a string slice. The second value is true if the item is found.
+func IndexOfString(slice []string, value string) (int, bool) {
```
tekante: This is probably due to general newness to go and it's structures and syntax but it took me a while to realize that this was testing for the presence of an element in a collection of elements rather than the presence of a substring within a string. Probably baggage from character arrays and the concept of a slice of a string from Perl combined with previous indexOf type functions. I don't think it has to change I just thought I'd mention it.

This is relevant because commands of the form of `util.Command("bash", "-c", fmt.Sprintf("echo 'nameserver %s' | sudo tee /etc/resolver/vm", bridgeIP)).Run()` on 178 of dns.go would fail this check. It's not really an issue for Macs because that one is guaranteed to be proceeded by a command which will already trigger the Password prompt and stop the spinner (line 175 of dns.go). I think it might affect Linux however as the configureLinuxResolver function has no such preceding command.

I've attached an example of what this looks like on Linux.

privilege-prompt-test-results-linux

tekante: One potential avenue around this would be to import the strings package and change the conditional on line 7 to something like
`		if elem == value || strings.Contains(elem, value + " ") {`

This appears to function properly for both mac and linux though it changes the nature of this function such that you'll probably want to rename it.

util/logger.go
grayside: I wouldn't expect it to. I need the log.Spin() to start the spinner and the log.Warning() for the real message... let me test just using the same message for both, since that would avoid coder confusion.

util/slices.go
grayside: There's a convention with exec.Cmd that the tokens all be individual array elements. I see that subshell situations might be different. Will adjust.

grayside:
In addition to addressing CJ's feedback, I took the following steps:

* Refactored Executor.ToString() to Executor.String() to be more in-line with Golang built-ins.
* Added the privilege escalation prompt to more of our command execution methods.
* When resuming the spinner, it now flashes "resuming operation", this is likely to be a flicker in many cases, but if the calling code doesn't do a good job of spinner management it may be confusing to see the spinning without context. I *think* this is a good idea, but could be convinced to leave it as an empty string.

febbraro:
I'm getting prompt overlap on `rig stop`

Using `sudo -k` first to remove privs

febbraro:
Prompt overlap is only when stopping a running machine. If I stop an already stopped machine, the prompt works fine.

grayside:
I haven't found a way to crack the rig stop spinner yet. It appears that the `route` command does something unusual in how it interacts with newlines or the password prompt, and may require much deeper research. Is this change a sufficient improvement to move forward anyway?

Should I output an extra message on rig stop to provide instruction on dealing with the confusion?

tekante:
I could go with proceeding if it is just the stop portion that overlap is wonky on. If it's something with route would an initial `sudo cat /dev/null` just so that the password gets prompted for alleviate that without causing ill effects?

grayside:
cat /dev/null seems to have helped. Will need to research the implications of that trick.

PR #141, opened 2018-02-01
Add spinner support to rig project sync[:stop]

grayside:
This is a first take, I think there's some language and formatting tightening potential, but it all seems to work as you'd expect.


commands/project_sync.go
```diff
@@ -119,20 +119,22 @@ func (cmd *ProjectSync) RunStart(ctx *cli.Context) error {
...
 	}
 
-	cmd.out.Channel.Info.Printf("Starting sync volume: %s", volumeName)
+	cmd.out.Info("Starting sync volume: %s", volumeName)
 	if err := util.Command("docker", "volume", "create", volumeName).Run(); err != nil {
 		return cmd.Failure(fmt.Sprintf("Failed to create sync volume: %s", volumeName), "VOLUME-CREATE-FAILED", 13)
 	}
-
-	cmd.out.Info("Starting Unison container")
+	cmd.out.Info("Sync volume '%s' created", volumeName)
```
tekante: I never see this message, even recording and playing in slow motion.

commands/project_sync.go
```diff
@@ -283,7 +286,7 @@ func (cmd *ProjectSync) LoadComposeFile() (*ComposeFile, error) {
 // when compiled without -cgo this executable will not use the native mac dns resolution
 // which is how we have configured dnsdock to provide names for containers.
 func (cmd *ProjectSync) WaitForUnisonContainer(containerName string, timeoutSeconds int) (string, error) {
-	cmd.out.Info("Waiting for container to start")
+	cmd.out.SpinWithVerbose("Sync container '%s' started , waiting for unison server process...", containerName)
```
tekante: I never see this message.

tekante: Ignore this, I wasn't supposed to because I was running in non verbose mode. It's there in verbose mode.

commands/project_sync.go
tekante: On further testing, I don't see it without verbose mode but I do see it in verbose mode.

tekante:
I dropped a few notes about where I don't see messages though I consider that not particularly blocking.

Here is the sequence I see:
Spinner: Starting Outrigger Filesync (unison)
Checkmark: Starting sync volume: nwpl-sync (overwrites prev spinner)
Spinner: Starting sync container: nwpl-sync (same name)
Checkmark: Sync container 'nwpl-sync' started (overwrites prev spinner)
Spinner: Initializing file sync...
Checkmark: Initial sync detected (overwrites prev spinner)
Spinner: Waiting for initial sync to finish
Checkmark: File sync completed (overwrites prev spinner)

tekante:
One other thing I noticed (which I also don't think needs to block) is the error message doesn't work as nicely as the success messages do. Since the message is getting returned however it doesn't seem to be as simple as swapping it to a logger message. Perhaps doing a log message and then returning the error?

rig-project-sync-error

tekante:
Adjusted error message to work better with spinner

rig-project-sync-error-split

PR #145, opened 2018-02-15
Fixed prune with a confirmation, a force, and removed the spinner

febbraro:


commands/prune.go
```diff
@@ -26,11 +26,16 @@ func (cmd *Prune) Commands() []cli.Command {
...
-		return cmd.Failure("Failure pruning Docker resources.", "COMMAND-ERROR", 13)
+
+	if util.AskYesNo("Are you sure you want to remove all unused containers, networks, images, caches, and volumes?") {
+		/* #nosec */
+		if exitCode := util.PassthruCommand(exec.Command("docker", "system", "prune", "--all", "--volumes", "--force")); exitCode != 0 {
+			return cmd.Failure("Failure pruning Docker resources.", "COMMAND-ERROR", 13)
+		}
+		cmd.out.Info("Unused Docker images, containers, volumes, and networks cleaned up.")
+	} else {
+		cmd.out.Info("Cleanup aborted.")
```
grayside: Should this be cmd.out.Warn()? (This is particularly useful if we run a spinner to differentiate no-action from successful action.)

commands/prune.go
```diff
@@ -26,11 +26,16 @@ func (cmd *Prune) Commands() []cli.Command {
...
 // Run executes the `rig prune` command
 func (cmd *Prune) Run(c *cli.Context) error {
-	cmd.out.Spin("Cleaning up unused Docker resources...")
-	/* #nosec */
-	if exitCode := util.PassthruCommand(exec.Command("docker", "system", "prune", "--all", "--volumes")); exitCode != 0 {
-		return cmd.Failure("Failure pruning Docker resources.", "COMMAND-ERROR", 13)
+
+	if util.AskYesNo("Are you sure you want to remove all unused containers, networks, images, caches, and volumes?") {
+		/* #nosec */
+		if exitCode := util.PassthruCommand(exec.Command("docker", "system", "prune", "--all", "--volumes", "--force")); exitCode != 0 {
```
grayside: Should we start a spinner inside here in case the command might take awhile to run?

tekante: If we kept the spinner I think it would still go above this conditional, just inside the AskYesNo block. I'm in favor of it since it can definitely take a while to run and there is no real indication that things are happening. If we leave it off then perhaps just mention in the prompt that this may take a while to execute.

febbraro: Well, the actual prune command has a lot of output as it is cleaning up things, so the spinner will likely get in the way. I'll test it out to see.

I'm thinking no spinner, but I'll see what it looks like.

tekante: I just did a test and I didn't see any output (so if some is being generated then the spinner is gobbling it). Interestingly, I didn't see the success message until I hit "Enter" while the spinner was running so I thought things might have actually gotten stuck. Interested to see if you get the same experience (and if so then perhaps the note that it'll will take a bit is a better course).

febbraro:
@grayside @tekante I added the `Warn` and the spinner.  The spinner worked for me, kinda, can you test it to see if it is something we would want to use?

febbraro:
https://www.dropbox.com/s/obm0lekorwssqjo/Screenshot%202018-02-22%2015.27.45.png?dl=0

tekante:
It worked for me though we still seem to have some weird conditions between when the spinner gets overwritten versus when it is augmented. See attached screenshot:

rig-prune

I don't know that those would be a big enough blocker to stop proceeding but it would be nice to figure out what is going on there (I think @grayside dug a good bit and this may be the best we can do).

I'll also note that I had to update to go 1.10 to build it as nfpm now depends on a Format element in archive/tar that 1.9 doesn't seem to recognize.

febbraro:
@grayside Do you have any thoughts here?  I'm torn if the intermittent appending vs newlines (see "Deleted Containers/Images" vs "Deleted Volumes") . Do we care about these inconsistencies? Is it worth it for the spinner? or should we disable the spinner and just drop a quick message like "This may take a while...."?

grayside:
How much wait time is there when nothing is being output to the screen? If it takes awhile but doesn't leave the user waiting to see progress, I think we should skip the spinner.

I find it jarring to mix the spinner with regular logs in the output for a single command execution, but prune is not a command we are ever likely to mix with something else, and if we do we can address it then.

PR #146, opened 2018-02-21
Shouty errors - Highly verbose guidance when errors happen

grayside:


febbraro:
I really liked this, do you think it is good to merge?

grayside:
If the code looks good, go for it! Is RIG_POWER_USER_MODE and --power-user the flag we want? I was trying to think of a "stop yapping at me" toggle.

febbraro:
I think that `power-user` mode is fine for now.  I'll have to see if the message bothers me.  I was thinking initially that this message should only be part of verbose mode, but I'm guessing you considered that already.

grayside:
Yeah, it seems like the people that most need help don't think about doctor or verbose, which is why that is output in this message. It's almost an argument to hide this in verbose mode :)

Another idea I had was to have it show the full message only 3 times for a given error name, or per day, or the like, doing some state tracking via a home directory dotfile.

febbraro:
I say lets get it in, give as much info as possible and dial it back as needed if it is too much trouble (how could it be).  Tools like `npm` are really loud when they fail.

PR #150, opened 2018-03-02
Developer Testing Commands

grayside:
Adds a couple simple developer testing commands. Very simple, guaranteed success/error state, that allows looking at how the spinner, logging, and notification works. We might add some more interaction cases to these in the future to facilitate manual/functional testing.

The hidden attribute means they are not displayed in help, and are only usable if you know they are there.


grayside:
Spinner finalization commit was for #149. It slipped in here, but please review as part of the other PR. I can clean this up if preferred.

PR #157, opened 2018-03-05
Refactor project scripts into cleaner API.

grayside:
Some refactoring to enable `rig project doctor`, replacing the existing long-standing PR later today I expect.

### DIFF
--- a/commands/project_eval.go
+++ b/commands/project_eval.go
@@ -10,6 +10,8 @@ import (
 	"github.com/phase2/rig/util"
 )
 
+// ProjectEval wrapps the evaluation of project scripts.
+// It mimics command struct except with unexported values.
 type ProjectEval struct {
 	out    *util.RigLogger
 	config *ProjectConfig
@@ -20,22 +22,22 @@ type ProjectEval struct {
 // Commands are run from the directory context of the project if available.
 // Use ProjectScriptRun to run a comman for potential user interaction.
 func (p *ProjectEval) ProjectScriptRun(script *ProjectScript, extra []string) int {
-	p.out.Verbose("Initializing project script '%s': %s", script.Id, script.Description)
+	p.out.Verbose("Initializing project script '%s': %s", script.ID, script.Description)
 	p.addCommandPath()
 	dir := p.GetWorkingDirectory()
 	shellCmd := p.GetCommand(script.Run, extra, dir)
-	p.out.Verbose("Evaluating Script '%s'", script.Id)
+	p.out.Verbose("Evaluating Script '%s'", script.ID)
 	return util.PassthruCommand(shellCmd)
 }
 
 // ProjectScriptResult matches ProjectScriptRun, but returns the data from the
 // command execution instead of "streaming" the result to the terminal.
 func (p *ProjectEval) ProjectScriptResult(script *ProjectScript, extra []string) (string, int, error) {
-	p.out.Verbose("Initializing project script '%s': %s", script.Id, script.Description)
+	p.out.Verbose("Initializing project script '%s': %s", script.ID, script.Description)
 	p.addCommandPath()
 	dir := p.GetWorkingDirectory()
 	shellCmd := p.GetCommand(script.Run, extra, dir)
-	p.out.Verbose("Evaluating Script '%s'", script.Id)
+	p.out.Verbose("Evaluating Script '%s'", script.ID)
 	return util.CaptureCommand(shellCmd)
 }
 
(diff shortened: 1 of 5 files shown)


grayside:
Heh, finding gometalinter started doing unpredictable stuff locally (scanning dependencies and so on). Will address these errors:

```
23.59s$ gometalinter --vendor --deadline=60s --config=gometalinter.json ./...
commands/project_eval.go:50::warning: Subprocess launching with variable.,HIGH,HIGH (gas)
commands/project_eval.go:52::warning: Subprocess launching with variable.,HIGH,HIGH (gas)
commands/project.go:85:16:warning: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)
commands/project_config.go:17:9:warning: struct field Id should be ID (golint)
commands/project_eval.go:13:6:warning: exported type ProjectEval should have comment or be unexported (golint)
```

commands/project_eval.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ProjectScriptResult matches ProjectScriptRun, but returns the data from the
+// command execution instead of "streaming" the result to the terminal.
+func (p *ProjectEval) ProjectScriptResult(script *ProjectScript, extra []string) (string, int, error) {
+	p.out.Verbose("Initializing project script '%s': %s", script.ID, script.Description)
+	p.addCommandPath()
+	dir := p.GetWorkingDirectory()
+	shellCmd := p.GetCommand(script.Run, extra, dir)
+	p.out.Verbose("Evaluating Script '%s'", script.ID)
+	return util.CaptureCommand(shellCmd)
```
febbraro: Since this is exactly the same as ProjectSriptRun, what is we passed in the output writer and if it is null Passthru, otherwise capture? (not sure it makes sense, but a thought)

util/shell_exec.go
```diff
@@ -57,7 +57,7 @@ func PassthruCommand(cmd *exec.Cmd) (exitCode int) {
 	cmd.Stdout = os.Stdout
 	cmd.Stdin = os.Stdin
 
-	bin := Executor{cmd}
+	bin := Convert(cmd)
```
febbraro: Where is Convert defined?

commands/project_eval.go
grayside: The thing is I want to capture stdout, stderr, and the exit code. The timing of when or if we output the result of the command depends on the caller. For an example of when this matters, see the project doctor PR which outputs the stdout verbosely, and may not even do that in the future.

util/shell_exec.go
grayside: It's in util. I don't remember when we added it but I figured if we already have a sort of "factory method" for this, might as well use it in more than one place.

commands/project_eval.go
```diff
@@ -0,0 +1,86 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/phase2/rig/util"
+)
+
+// ProjectEval wrapps the evaluation of project scripts.
+// It mimics command struct except with unexported values.
+type ProjectEval struct {
```
febbraro: Would a better name be `ProjectScript` since it actually wraps the script/execution then you can just have a method of `Run()` and `Result()`?

grayside: We already have ProjectScript and it's part of how the project_config.go unmarshals the yaml file. We could move ProjectScript to this file, even rename the file to match, and leave project_config.go to reference the struct from the new one.

febbraro: Yeah that sounds good to me, lets move the `ProjectScript` struct here and add these methods to it because they represent the same thing, the config of a project script and the methods to execute it.

febbraro: An alternative is to just call the config struct `Script`  like we did with `Sync` and then rename this to `ProjectScript` just like we have `ProjectSync`

grayside:
I took the Script/ProjectScript approach, there is a need for two different structs: one of them is initialized with project config and the logger and operates as a "runner" service for a Script. This ended up be a fairly minimal change.

commands/project_script.go
```diff
@@ -0,0 +1,88 @@
...
+// the project script and bonus arguments from the extra parameter.
+// Commands are run from the directory context of the project if available.
+// This also supports follow-up user interaction.
+func (p *ProjectScript) Run(script *Script, extra []string) int {
+	p.out.Verbose("Initializing project script '%s': %s", script.ID, script.Description)
+	p.addCommandPath()
+	dir := p.GetWorkingDirectory()
+	shellCmd := p.GetCommand(script.Run, extra, dir)
+	shellCmd.Env = append(os.Environ(), "RIG_POWER_USER_MODE=1")
+	p.out.Verbose("Evaluating Script '%s'", script.ID)
```
febbraro: Seems like the previous 6 lines are the same as in `Capture` and should be DRY'd to something like

`util.PassthruCommand(p.GenerateCommand())` or something like it.

PR #158, opened 2018-03-05
Project Doctor

grayside:
Here's a reboot on the project doctor work. This is based on #157 and should be reviewed after that.

With some tweaks to the built-in conditions we could merge this as-is, but there are a few things I'd like to address:

### Uncomfortable Things

* Extract the two defined conditions into working examples. I found it really difficult to parameterize for container running checks, but if we go with the sync container name in the outrigger.example.yml it will work fine.

### Future Enhancements

* Add a `rig project sync:name` command to supply the sync volume and sync container name.
* Whether as rig commands or as gotemplate macros, it's really tricky to script good checks for whether container is running. I'd like to facilitate that so projects can add checks that their required services are live.
* Integrate with a future `rig project sync:doctor` command which verifies the health of the sync process. (I'm planning to tackle that to split the current checking rig project sync does as part of setup into a separately callable API method and command.)
* Concurrently execution doctor checks. This requires some stuff:
  * Addressing timeouts
  * Not outputting the results of the condition execution even in verbose mode as interleaving will be too confusing.
  * Potentially adding something like `rig project doctor --analyze=conditionId` to support testing that a condition behaves in isolation.

### DIFF
--- a/commands/project_doctor.go
+++ b/commands/project_doctor.go
@@ -1,0 +1,215 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/urfave/cli"
+)
+
+type ProjectDoctor struct {
+	BaseCommand
+	Config *ProjectConfig
+}
+
+// @TODO add an error tier for fatal, wherein analysis should be halted.
+const (
+	ConditionSeverityINFO    string = "info"
+	ConditionSeverityWARNING string = "warning"
+	ConditionSeverityERROR   string = "error"
+)
+
+type Condition struct {
+	Id           string
+	Name         string
+	Test         []string
+	Diagnosis    string
+	Prescription string
+	Severity     string
+}
+
+type ConditionCollection map[string]*Condition
+
+func (cmd *ProjectDoctor) Commands() []cli.Command {
+	cmd.Config = NewProjectConfig()
+
+	diagnose := cli.Command{
+		Name:        "doctor:diagnose",
+		Aliases:     []string{"doctor"},
+		Usage:       "Run to evaluate project-level environment problems.",
+		Description: "This command validates known problems with the project environment. The rules can be extended via the 'doctor' section of the project configuration.",
+		Before:      cmd.Before,
+		Action:      cmd.RunAnalysis,
+	}
+
+	compendium := cli.Command{
+		Name:        "doctor:conditions",
+		Aliases:     []string{"doctor:list"},
+		Usage:       "Learn all the rules applied by the doctor:diagnose command.",
+		Description: "Display all the conditions for which the doctor:diagnose command will check.",
+		Before:      cmd.Before,
+		Action:      cmd.RunCompendium,
+	}
+
+	return []cli.Command{diagnose, compendium}
+}
+
+// RunAnalysis controls the doctor/diagnosis process.
+func (cmd *ProjectDoctor) RunAnalysis(ctx *cli.Context) error {
+	compendium, _ := cmd.GetConditionCollection()
+	if err := cmd.AnalyzeConditionList(compendium); err != nil {
+		// Directly returning the framework error to skip the expanded help.
+		// A failing state is self-descriptive.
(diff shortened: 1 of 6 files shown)


grayside:
* [ ] Rebase onto changes in the Project Scripts API branch
* [ ] See if we can merge more of the Condition struct particulars into the Script object. E.g., generalized forms of Diagnosis, Prescription, and Severity.
  * This will allow simplifying the code in project_doctor a little and move more handling to project_script.go
  * Secondarily, it will allow project scripts to be configured to have a success-but-warning mode, and have error and corrective messages for better user communication about next steps.
  * Third, while the naming conventions are "cute", standardizing naming might be more consistent & learnable.
  * Doing this will probably require a Condition struct to alias or shadow the Script struct

commands/project_doctor.go
```diff
@@ -0,0 +1,213 @@
...
+		Severity:     ConditionSeverityWARNING,
+	}
+	if _, ok := conditions["sync-container-not-running"]; !ok {
+		conditions["sync-container-not-running"] = item1
+	}
+
+	item2 := &Condition{
+		Id:           "sync-volume-missing",
+		Name:         "Sync Volume is Missing",
+		Test:         []string{fmt.Sprintf("$(id=$(docker container ps -aq --filter 'name=^/%s$'); docker top $id &>/dev/null)", syncName)},
```
tekante: Is this supposed to be the exact same test as the one that checks to see if the container is running?

commands/project_doctor.go
```diff
@@ -0,0 +1,213 @@
...
+		Diagnosis:    "The Sync container for this project is not available.",
+		Prescription: "Run 'rig project sync:start' before beginning work. This command may be included in other project-specific tasks.",
+		Severity:     ConditionSeverityWARNING,
+	}
+	if _, ok := conditions["sync-container-not-running"]; !ok {
+		conditions["sync-container-not-running"] = item1
+	}
+
+	item2 := &Condition{
```
tekante: Should we also add a test to the effect of `ps aux | grep unison | $syncName` to check for the process on the local machine?

commands/project_doctor.go
tekante: Perhaps it should be `docker volume ls $syncName` which is pretty easy to do as a shell test (assuming the volume name would get hard coded in outrigger.yml).

commands/project_doctor.go
```diff
@@ -0,0 +1,213 @@
...
+	// @TODO move these to outrigger.yml once we have pure shell facilities.
+	eval := ProjectEval{cmd.out, cmd.Config}
+	sync := ProjectSync{}
+	syncName := sync.GetVolumeName(cmd.Config, eval.GetWorkingDirectory())
+
+	// @todo we should have a way to determine if the project wants to use sync.
+	item1 := &Condition{
+		Id:           "sync-container-not-running",
+		Name:         "Sync Container Not Working",
+		Test:         []string{fmt.Sprintf("$(id=$(docker container ps -aq --filter 'name=^/%s$'); docker top $id &>/dev/null)", syncName)},
```
tekante: Some potential shell versions assuming you are ok with the sync volume being hard coded in outrigger.yml.

`test -z $(docker container ps -aq --filter 'name=^/projectname-sync$')` should tell you if the container exists but not much about if it is running.

`test -z $(docker container ps -q --filter 'name=^/projectname-sync$')` dropping the -a flag should let you know if it is running but can't tell you anything about the processes inside it. I don't think that is different than what the return code from docker top is doing for you but I'm not positive.

commands/project_doctor.go
grayside: I'm working on a PR which adds a `rig project sync:name` to get that.

commands/project_doctor.go
grayside: The problem is `docker container ps` never has a failing exit code, whereas `docker top` does.

After creating these examples of conditions, I've since gone on to work on have a more hard-coded set of checks around project sync, I hope to have a PR ready later today/tomorrow.

With all the issues around unison, I want to be able to tell people to upgrade rig and have a comprehensive solution in place for all projects using unison.

PR #159, opened 2018-03-05
Create 'rig ssh' command

dnmurray:


commands/ssh.go
```diff
@@ -0,0 +1,43 @@
...
+	BaseCommand
+}
+
+// Commands returns the operations supported by this command
+func (cmd *Ssh) Commands() []cli.Command {
+	return []cli.Command{
+		{
+			Name:  "ssh",
+			Usage: "Start an ssh session into the docker-machine vm",
+			Flags: []cli.Flag{
```
febbraro: With no flags you can remove this.

grayside: Can simply remove the flags if none exist.

commands/ssh.go
```diff
@@ -0,0 +1,43 @@
...
+			Action: cmd.Run,
+		},
+	}
+}
+
+// Run executes the `rig start` command
+func (cmd *Ssh) Run(c *cli.Context) error {
+	// Does the docker-machine exist
+	if !cmd.machine.Exists() {
+		return fmt.Errorf("docker machine %s not found", cmd.machine.Name)
```
febbraro: We have a different approach to error reporting now. See start or remove or something. There will be an error code and key.

grayside: This works fine to return the error, but if you want the notification you would return cmd.Failure(). I kind of like the simpler error for this use case.

commands/ssh.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	exitCode := util.PassthruCommand(exec.Command("docker-machine", "ssh", cmd.machine.Name))
```
grayside: Nice job finding the right way to invoke this for an interactive session! We don't make it terribly obvious.

commands/ssh.go
```diff
@@ -0,0 +1,43 @@
...
+	if exitCode == 0 {
+		return nil
+	} else {
```
grayside: No need for an `else` here, in fact our linter will complain.

commands/ssh.go
febbraro: With Go you can combine this command and the boolean evaluation to have just a then dealing with error or otherwise, then just return Nil (or more specifically `return cmd.Success()`)

commands/ssh.go
```diff
@@ -0,0 +1,43 @@
...
+		return fmt.Errorf("Exit code was %d", exitCode)
```
febbraro: Look at the `Project.run` command for how we error/code/message for exec commands that failed.

grayside: If you instead do:
```
cli.NewExitError(fmt.Sprintf("ERROR: %s [%s] (%d)", message, errorName, exitCode)
```

It will match our short error code and exit rig with the same code the ssh command exits on.

commands/ssh.go
grayside: If you append something like strings.Join(ctx.Args(), " ") you can allow for `rig ssh ls -l`.

febbraro: Great suggestion

commands/ssh.go
febbraro: `cmd.Error` handles that, I thought.

commands/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Run executes the `rig ssh` command
+func (cmd *Ssh) Run(c *cli.Context) error {
+	// Does the docker-machine exist
+	if !cmd.machine.Exists() {
+		return fmt.Errorf("docker machine %s not found", cmd.machine.Name)
+	}
+
+	if exitCode := util.PassthruCommand(exec.Command("docker-machine", "ssh", cmd.machine.Name)); exitCode == 0 {
+		return cmd.Success("")
```
febbraro: I would make this as the default return for the function (outside of an if)

commands/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+	} else {
+		return cmd.Failure(fmt.Sprint("Failure running 'docker-machine ssh %s'", cmd.machine.Name), "COMMAND-ERROR", exitCode)
```
febbraro: I woudl change the if to have `exitCode != 0` and perform the error there, then you dont need an else and can just return the success.

commands/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+			Action: cmd.Run,
+		},
+	}
+}
+
+// Run executes the `rig ssh` command
+func (cmd *Ssh) Run(c *cli.Context) error {
+	// Does the docker-machine exist
+	if !cmd.machine.Exists() {
+		return fmt.Errorf("docker machine %s not found", cmd.machine.Name)
```
febbraro: Use `cmd.Error` here like below

febbraro:
This is awesome, thanks @dnmurray

dnmurray:
@grayside how would I do a `cd /data` after the session starts? I used `PassthruCommand()` to get stdin passed through to the command so the user stays in the ssh session. I don't know how I could pass in a command to execute and then return control to the user's stdin. `docker-machine ssh cd /data` just returns to the OSX prompt when I test it manually (run one command and exit the ssh session).

grayside:
Well, I see how to run a command, but not how to change directory inside: https://docs.docker.com/machine/reference/ssh/

```
docker-machine ssh dev ls /data
```

febbraro:
Awesome, thanks Dave!

PR #162, opened 2018-03-07
Activate power user mode when running rig project scripts.

### LINKED ISSUE #161: When chaining rig project scripts "shouty errors" are redundantly output
When you configure a chain of rig project scripts several layers deep, and the bottom-most layer fails, this will trigger failures up the stack. As a result, you will get multiple "shouty" outputs as each project script fails.

Solution: When rig-project script is invoked, have it set the power users environment variable such that only the outer-most script "shouts".

This will require taking care to merge the existing env vars from the environment with this new variable to avoid causing a regression.

grayside:
This prevents the recursion of getting one shouty error per rig project run:* instance in a chain of commands.

Fixes #161 

Needs to be ported into #157

### DIFF
--- a/commands/project.go
+++ b/commands/project.go
@@ -94,6 +94,7 @@ func (cmd *Project) Run(c *cli.Context) error {
 
 		shellCmd := cmd.GetCommand(scriptCommands)
 		shellCmd.Dir = dir
+		shellCmd.Env = append(os.Environ(), "RIG_POWER_USER_MODE=1")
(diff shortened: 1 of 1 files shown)


febbraro:
You said it needs to be ported into #157, do you just want to add it there and close this PR?

grayside:
If we can get that merged and released in the next few days/week. This is really sloppy right now for affected projects. Depends on the feedback over there :)

PR #170, opened 2018-06-10
Have rig check for any updates available to itself

dnmurray:
Have rig check it's version against the latest release during the `start` and `doctor` commands. Say something if there's an update available.

### DIFF
--- a/commands/doctor.go
+++ b/commands/doctor.go
@@ -206,5 +206,12 @@ func (cmd *Doctor) Run(c *cli.Context) error {
 		}
 	}
 
+	// 7. Check for availability of a rig upgrade
+	cmd.out.Spin("Checking for available rig updates...")
+	if msg := util.CheckForRigUpdate(c.App.Version); msg != "" {
+		cmd.out.Info(msg)
+	} else {
+		cmd.out.Info("rig is up-to-date")
+	}
 	return nil
 }
--- a/commands/doctor.go
+++ b/commands/doctor.go
@@ -210,6 +210,8 @@ func (cmd *Doctor) Run(c *cli.Context) error {
 	cmd.out.Spin("Checking for available rig updates...")
 	if msg := util.CheckForRigUpdate(c.App.Version); msg != "" {
 		cmd.out.Info(msg)
+	} else {
+		cmd.out.Info("rig is up-to-date")
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


commands/doctor.go
```diff
@@ -206,5 +206,10 @@ func (cmd *Doctor) Run(c *cli.Context) error {
 		}
 	}
 
+	// 7. Check for availability of a rig upgrade
+	cmd.out.Spin("Checking for available rig updates...")
+	if msg := util.CheckForRigUpdate(c.App.Version); msg != "" {
+		cmd.out.Info(msg)
+	}
```
febbraro: I thought it was nice that doctor would say that `rig` was up to date. You can leave that message, if you'd like

util/update.go
```diff
@@ -0,0 +1,92 @@
...
+	// Local dev, version == "master" which isn't going to parse.
+	curVer, verr := version.NewVersion(curRigVersion)
+	if tag, err := currentRigReleaseTag(); err == nil {
+		if tagVer, verr2 := version.NewVersion(tag); verr2 == nil {
+			if verr != nil || tagVer.Compare(curVer) > 0 {
+				return "An update for rig is available: " + tag
+			}
+		}
+	} else {
+		if Logger().IsVerbose {
```
febbraro: If you just use `Logger().Verbose(...)` you don't have to do a verbose check.  In fact, it's incorrect to hide a warning behind verbose, verbose it literally for extra noisy messages, a warning does not categorize as chatty.

dnmurray: Thanks. I was on the fence about reporting errors/warning in general and took the approach "if you want to see them, put in -v." I expected most people wouldn't care that the update check didn't work (so I was just failing silently). Trying to reduce distractions/support issues for what would be a bug in the program.

util/update.go
```diff
@@ -0,0 +1,79 @@
...
+	// Local dev, version == "master" which isn't going to parse.
+	curVer, verr := version.NewVersion(curRigVersion)
+	if tag, err := currentRigReleaseTag(); err == nil {
+		if tagVer, verr2 := version.NewVersion(tag); verr2 == nil {
+			if verr != nil || tagVer.Compare(curVer) > 0 {
+				return "An update for rig is available: " + tag
+			}
+		}
+	} else {
+		// Local dev, this always appears, since version get set by the build.
```
tekante: I think this comment is meant to apply to line 24? When I was testing locally I only saw this message if there was trouble talking to github to get the current version.

If it does apply to line 24 I think we can avoid it printing during local development by changing the condition to `if verr == nil && tagVer.Compare...`

PR #172, opened 2018-06-10
Feature/save dns records to hosts

dnmurray:
Unix only so far.  Need to do some research to support Windows (and someone to test).


commands/dns-records.go
```diff
@@ -31,27 +82,31 @@ func (cmd *DNSRecords) Commands() []cli.Command {
...
-			// attach any aliases too
-			for _, a := range record["Aliases"].([]interface{}) {
-				fmt.Printf("%s\t%s\n", ip, a)
-			}
-		}
+	if c.Bool("save") {
+		return cmd.saveDNSRecords(records)
 	}
 
+	printDNSRecords(records)
```
febbraro: if you do a `save` or `remove` you likely don't want the records printed out too so maybe we only print the records if `save` or `remove` have not been specified

commands/dns-records.go
```diff
@@ -74,11 +129,142 @@ func (cmd *DNSRecords) LoadRecords() ([]map[string]interface{}, error) {
...
+
+// ⚠ Administrative privileges needed...
+
+func (cmd *DNSRecords) saveDNSRecordsUnix(records []*DNSRecord) error {
+	// Both of these are []string
+	oldHostEntries := util.LoadFile("/etc/hosts")
+	newHostEntries := stripDNS(oldHostEntries)
+	// records.String does the formatting, so convert both to a string
+	oldHosts := strings.Join(oldHostEntries, "\n")
+	newHosts := strings.Join(newHostEntries, "\n") + "\n" +
```
febbraro: You could also do something like this...

```
fmt.Sprintf(`%s
%s
%s`, x, y, z)
```

to get the newlines you want.

febbraro:
The other thing that comes to mind here is how many people may get hung up because they have entries in their hosts files that they haven't removed. Maybe add a doctor check for DNS entries in `/etc/hosts` and let the user know that?  Or do you think this is really a non issue?

commands/dns-records.go
dnmurray: They always return if they do a save or remove.  So it just falls through to the default if they don't specify either.

dnmurray:
I can add a check that the rows are there.  It's easy enough to look for, since I'm using the pre/postamble.  If you do a --save and the rows are already they, they get removed and recreated.

commands/dns-records.go
```diff
@@ -17,12 +19,61 @@ type DNSRecords struct {
...
+	IPs     []string
+	TTL     int64
+	Aliases []string
+}
+
+// DNSRecordsList is an array of DNSRecords
+type DNSRecordsList []*DNSRecord
+
+const (
+	unixHostsPreamble  = "##+++ added by rig"
```
tekante: I think you'll use this regardless of the underlying os as markers for starting/removing items so perhaps just call this hostsPreamble

commands/dns-records.go
```diff
@@ -74,11 +129,142 @@ func (cmd *DNSRecords) LoadRecords() ([]map[string]interface{}, error) {
 	}
 
 	dnsdockMap, _ := js.Map()
-	records := []map[string]interface{}{}
-	for id, value := range dnsdockMap {
-		record := value.(map[string]interface{})
-		record["Id"] = id
+	records := make([]*DNSRecord, 0, 20)
```
tekante: Interested in your thoughts behind the use of the capacity argument here. As near as I can tell this doesn't set any upper bound on the size of the array though I had to write a quick test to make sure. Is the initial capacity you picked arbitrary (both here and below where you used 10) or is there something behind that particular value?

dnmurray: Completely arbitrary. Yes, when accumulating the array contents via `append()` more space will be allocated if needed.

commands/dns-records.go
dnmurray: Good point.  Thanks.

dnmurray:
Can this be tagged "In Progress".  Adding Windows support.

commands/dns-records.go
tekante: Given that it's a dynamic array perhaps just drop the capacity argument here and anywhere it doesn't have an obvious purpose. This is to prevent confusing anyone in the future about why it is there, how it got picked, does it actually result in enforced bounds, etc? Alternatively, comment something to the effect of what I think may be true about it:

// capacity picked arbitrarily with a size likely to prevent need to reallocate underlying arrays while appending items in most cases

I also presume that has some potential performance benefit though at the scales here I think it's unobservable.

If that comment isn't accurate I'd be interested in what the benefits are of declaring a capacity are in the case where append is going to be used. When using copy it's an obvious useful item to restrict the number of items you want from a potentially larger array.

Thanks for the pointers as I drag myself deeper into go.

commands/dns-records.go
febbraro: OH yeah. ha. You got it.

febbraro:
This has a bunch of merges in it. Do you need to rebase on `develop`?

PR #173, opened 2018-06-19
dep ensure --update to pick up a new gospinner and other things

dnmurray:
gospinner has a new release tag, to address https://github.com/phase2/rig/issues/160.  Did a `dep ensure --update` and picked up that and a few other changes.  This should be part of the next RC so we can get some testing done.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -16,14 +16,14 @@
 [[projects]]
   name = "github.com/fatih/color"
   packages = ["."]
-  revision = "507f6050b8568533fb3f5504de8e5205fa62a114"
-  version = "v1.6.0"
+  revision = "5b77d2a35fb0ede96d138fc9a99f5c9b6aef11b4"
+  version = "v1.7.0"
 
 [[projects]]
   branch = "master"
   name = "github.com/hashicorp/go-version"
   packages = ["."]
-  revision = "4fe82ae3040f80a03d04d2cccb5606a626b8e1ee"
+  revision = "23480c0665776210b5fbbac6eaaee40e3e6a96b7"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


febbraro:
I see a lock file was updated, but what about the `Gopkg.toml` file?

dnmurray:
`dep ensure --update` didn't make any changes, and I didn't make any changes.  I didn't delve real deep in how semantic version matches are happening, but it did pick up the update to gospinner and the other packages, without me changing the .toml file. On current master:

```
rig$ dep ensure -update
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/mattn/go-isatty

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies,

rig$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   Gopkg.lock

no changes added to commit (use "git add" and/or "git commit -a")
```

febbraro:
What about if you delete the `vendor` directory and do a `dep ensure` does it capture the correct version?  I'm guessing a clean install may not, so we would need to update the `.toml` file too

dnmurray:
Verified that we needed to update the .toml file (I created a fresh install to test, and `dep ensure` just brought down the pinned packages). yaml.v2 got updated in the original `dep ensure` but since the .toml file specifies a branch and not a version, I didn't change that.

PR #176, opened 2018-08-11
Update of changelog for 2.2.0

tekante:


dnmurray:
Did the linter get updated? Maybe we need to change the rules to be a little less restrictive for what it's throwing now.

